PR #18, opened 2016-09-12
fix reset link in mail if running self-service-password with differen…

TheHendla:
…t context path


coudot:
Seems this was not mandatory. Changes canceled in #30

PR #28, opened 2016-09-26
Updated make_ad_password

med-amine-21:
Modified make_ad_password so that it now accepts accentued characters. 
The modification uses mbstring which is already used in other functions so there is no problem with additional dependecies.

### DIFF
--- a/lib/functions.inc.php
+++ b/lib/functions.inc.php
@@ -84,7 +84,7 @@ function make_md4_password($password) {
 # Create AD password (Microsoft Active Directory password format)
 function make_ad_password($password) {
     $password = "\"" . $password . "\"";
-    $adpassword=mb_convert_encoding($password, "UTF-16LE","UTF-8");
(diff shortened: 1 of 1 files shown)


lib/functions.inc.php
```diff
@@ -84,11 +84,7 @@ function make_md4_password($password) {
 # Create AD password (Microsoft Active Directory password format)
 function make_ad_password($password) {
     $password = "\"" . $password . "\"";
-    $len = strlen(utf8_decode($password));
-    $adpassword = "";
-    for ($i = 0; $i < $len; $i++){
-        $adpassword .= "{$password{$i}}\000";
-    }
+    $adpassword=mb_convert_encoding($password, "UTF-16LE","UTF-8");
```
coudot: Ok with the patch, but please respect some coding rules by adding some spaces:
`$adpassword = mb_convert_encoding($password, "UTF-16LE", "UTF-8");`

Please resubmit after this little change so I can merge your code.

med-amine-21:
I've commited the change this  time respecting spacing rules

PR #55, opened 2016-11-23
shadowExpire in LDAP

deniskin82:
allow to change/reset shadowExpire in LDAP


coudot:
Hello,

thanks for the patch.

You choose to force the value to -1, which means the password will never expire. Maybe we could have the expiration time as configuration parameter, and update shadowExpire with the correct value when changing the password?

deniskin82:
Hi,
I came up with something. What do u think?

coudot:
Seems good, thanks!

PR #62, opened 2016-12-13
Updated german translations + added new features

dlang9159:
Hello everyone,

I recently installed your project on my server and made some changes for my specific application. Now I wanted to share my changes to the public.

1. Updated the german translation, now every string should be available in german. (file is now in utf-8, on my server I got no problem with that)
2. Added a $mail_signature in config.inc.php, this string is attached to every mail message. Empty by default
3. Added a feature to use the user mail address from ldap directory for "requestbytoken" instead of user provides mail address. This is enabled by "$mail_address_use_ldap = true;" disabled by default.


coudot:
Thanks a lot, but would it be possible to do 3 different PR?

dlang9159:
I'm sorry, of course I will create separate pull requests

PR #64, opened 2016-12-14
Mail from ldap

dlang9159:
Added a feature to use the user mail address from ldap directory for "requestbytoken" instead of user provides mail address. This is enabled by "$mail_address_use_ldap = true;" disabled by default.

(Please ignore the additional commits, those were reverted and do not affect anything. I messed up my repository a little bit)


dlang9159:
Will reopen new pull request with cleaned up commits. Sorry for the inconvenience, I'm still getting used to github.

PR #65, opened 2016-12-14
Mail signature

dlang9159:
Added a $mail_signature in config.inc.php, this string is attached to every mail message. Empty by default

(Please ignore the additional commits, those were reverted and do not affect anything. I messed up my repository a little bit)


dlang9159:
Will reopen new pull request with cleaned up commits. Sorry for the inconvenience, I'm still getting used to github.

PR #71, opened 2017-01-05
Mail use fullname

dlang9159:
I wrote this because I wanted my emails to look more pretty by displaying the fullname of a user instead of the username/login. Instead of "jsmith92" you can also use "Jones Smith" or just "Jones". Default option is still the login name.

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -30,6 +30,7 @@ $ldap_bindpw = "secret";
 $ldap_base = "dc=example,dc=com";
 $ldap_login_attribute = "uid";
 $ldap_fullname_attribute = "cn";
+$ldap_givenname_attribue = "givenName";
 $ldap_filter = "(&(objectClass=person)($ldap_login_attribute={login}))";
 
 # Active Directory mode
@@ -154,6 +155,11 @@ $mail_address_use_ldap = false;
 $mail_from = "admin@example.com";
 $mail_from_name = "Self Service Password";
 $mail_signature = "";
+# Which Anrede is used in mails?
+# login (default)
+# fullname
+# givenname
+$mail_display_name = "login";
 # Notify users anytime their password is changed
 $notify_on_change = false;
 # PHPMailer configuration (see https://github.com/PHPMailer/PHPMailer)
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -155,7 +155,7 @@ $mail_address_use_ldap = false;
 $mail_from = "admin@example.com";
 $mail_from_name = "Self Service Password";
 $mail_signature = "";
-# Which Anrede is used in mails?
+# Which display name is used in mails?
 # login (default)
 # fullname
 # givenname
--- a/pages/change.php
+++ b/pages/change.php
@@ -32,6 +32,7 @@ $newpassword = "";
 $oldpassword = "";
 $ldap = "";
 $userdn = "";
+$display_name_lookup = "";
 if (!isset($pwd_forbidden_chars)) { $pwd_forbidden_chars=""; }
 $mail = "";
 
@@ -129,6 +130,9 @@ if ( $result === "" ) {
         if ( $mailValues["count"] > 0 ) {
             $mail = $mailValues[0];
         }
+        $display_name_lookup["login"] = ldap_get_values($ldap, $entry, $ldap_login_attribute);
+        $display_name_lookup["givenname"] = ldap_get_values($ldap, $entry, $ldap_givenname_attribute);
+        $display_name_lookup["fullname"] = ldap_get_values($ldap, $entry, $ldap_fullname_attribute);       
     }
 
     # Check objectClass to allow samba and shadow updates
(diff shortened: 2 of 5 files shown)


dlang9159:
Found a bug, will resolve it by tomorrow.

coudot:
Thanks for the PR, will have a look later.

dlang9159:
What is the status here?

plewin:
I believe SSP should provide a way to override lang files without touching the provided files in /lang/* to offer more flexibility.

For example, by including a new file "conf/lang_overrides.inc.php" after the `require_once("lang/$lang.inc.php");`
https://github.com/ltb-project/self-service-password/blob/master/index.php#L47

This file would replace {login} by {firstname} in email templates, $messages['resetmessage'], $messages['changemessage']

The 
`$data = array( "login" => $display_name_lookup[$mail_display_name], "mail" => $mail, "url" => $reset_url ) ;`
would be
`$data = array( "login" => $login, "mail" => $mail, "url" => $reset_url, "firstname" => ..., ..., ..) ;`

dlang9159:
@plewin thanks for your idea, this would be a much better way to deal with it.
In this case it would be great to update the whole language "backend" a little bit. The current way of dealing with translations seems a bit cumbersome to me.

plewin:
@dlang9159 Sorry if it is getting a bit off topic but what changes would you like to see in the language backend ?

dlang9159:
@plewin For instance the one you already mentioned: I'd like to alter some strings like the mail text for example add a signature ( #67 was kind of a workaround).

But the most "annoying" thing is that you have to add new language strings to all *.inc.php files manually (or did I miss something). It would be more convenient to use a look-up function that returns the (english) default string if the language doesn't support it yet.

But this is just my personal opninion and it is open to discussion. I am aware that this leads to more complexity.

Maybe I should add an issue where we can discuss this in more depth.

coudot:
Sorry for delay, I still need to check the code and merge it

coudot:
Changing how we manage language files will not solve completely this issue, as we need to search in LDAP directory the attribute we want to put in the message.

So we need at least a configuration parameter that can be an array and set all attributes we want to get from user entry.

Then we can indeed have an easy way to override lang messages. I've done that in White Pages:
```php
if (file_exists("../conf/$lang.inc.php")) {
    require_once("../conf/$lang.inc.php");
}
```

It allows to define a lang file in conf directory that will override or add messages.

What do you think?

dlang9159:
This looks ok for me. Of course we must define what attributes we want to get.
I suggest: uid/login, givenName, fullname. 
Any other attributes? How do we handle attributes with multiple values?

If we agree on your suggestion for the language file I will alter my PR to support this feature.

coudot:
@dlang9159 Attributes will be configured, but we can indeed have a default value with these ones, for example:
````php
$ldap_attributes = array('uid', 'cn', 'sn', 'givenname');
````

I will open a new issue for the language file override.

**Comment by plewin** [comment]
I believe the default should be an empty array

```
# Fetch extra attributes for the email templates
# Example : $fetch_extra_attributes_for_mail = array('uid', 'cn', 'sn', 'givenname');
# Template will be able to use {uid} {cn} {sn} {givenname}
$fetch_extra_attributes_for_mail = array();
```

If there is no extra attributes to fetch, nothing will fetched.

Also I think there is a bit too much of code duplication in this PR.

dlang9159:
@plewin you are totally right with code duplication, I will move most of it to lib/functions.inc.php

gino909:
Hi 
to make this happen I have modified following scripts
pages/sendtoken.php

`# Match with user submitted values
    foreach ($mailValues as $mailValue) {
        if (strcasecmp($mail_attribute, "proxyAddresses") == 0) {
            $mailValue = str_ireplace("smtp:", "", $mailValue);
        }
        if (strcasecmp($mail, $mailValue) == 0) {
            $match = 1;
            **# Get user name for notification
            $cnValues = ldap_get_values($ldap, $entry, $ldap_fullname_attribute);
            $username = $cnValues[0];**
        }
    }
`
...
`
$data = array( "username" => $username, "login" => $login, "mail" => $mail, "url" => $reset_url ) ;
`

pages/resetbytoken.php

`
    # Get user email for notification
    if ( $notify_on_change ) {
        $mailValues = ldap_get_values($ldap, $entry, $mail_attribute);
        if ( $mailValues["count"] > 0 ) {
            $mail = $mailValues[0];
            **# Get user name for notification
            $cnValues = ldap_get_values($ldap, $entry, $ldap_fullname_attribute);
            $username = $cnValues[0];**
        }
    }

`
....
`
$data = array( "username" => $username, "login" => $login, "mail" => $mail, "password" => $newpassword);
`

pages/change.php

`
    # Get user email for notification
    if ( $notify_on_change ) {
        $mailValues = ldap_get_values($ldap, $entry, $mail_attribute);
        if ( $mailValues["count"] > 0 ) {
            $mail = $mailValues[0];
            **# Get user name for notification
            $cnValues = ldap_get_values($ldap, $entry, $ldap_fullname_attribute);
            $username = $cnValues[0];**
        }
    }

`
...
`
$data = array( "username" => $username, "login" => $login, "mail" => $mail, "password" => $newpassword);
`

I was not tested with Windows AD!

coudot:
We must find a cleaner way to get attributes from user and use them in mails. The configuration must be generic

derhelge:
+1

JuJuMDX:
Hello,

This is an old PR for SSP 1.1 so i would like to know if it's still necessary to keep it open ?

PR #72, opened 2017-01-05
Copying configuration file is now required

knixeur:
To avoid accidental commits of the configuration file, it was
removed in favour of a sample config file which needs to be copied
to be used as default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+config/config.inc.php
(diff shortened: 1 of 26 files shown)


knixeur:
Feel free to close if you don't see a point here.
In my case, I have to do some customization (which I'll later PR too) asked by security team and I don't want them to pull my config file and have conflicts with theirs, etc.

coudot:
Hi,

thanks for the suggestion, but it would require more work as this file is tagged as configuration file in our RPM/Debian packaging.

PR #81, opened 2017-02-10
Allow self service of sshPublicKey attribute in LDAP

praymann:
At my current workplace, we desire this functionality so I forked and coded it as a proof of concept.

Most of the code was there I just needed to copy/tweak as needed. There is one new action created called "changesshkey".

All the properties are there, and are false by default:
* change_sshkey (to enable the functionality)
* who_change_sshkey (who changes the attribute)
* notify_on_sshkey_change (notify users when attribute is changed)

I am not a PHP developer, so if this needs edits or changes let me know. Right now it's lacking translations into other languages.

If this is not desired functionality let me know or deny the pull request.


coudot:
Thanks for the PR. You can check the translations files to let Travis become green.

praymann:
Updated to include the translations so Travis would go green but it appears to be bombing for PHP 7.0 but not due to my changes.

```
$ phpunit tests
PHP Fatal error:  Class 'PHPUnit_Framework_TestCase' not found in /home/travis/build/ltb-project/self-service-password/tests/CheckPasswordTest.php on line 2
PHP Stack trace:
PHP   1. {main}() /home/travis/.phpenv/versions/7.0.16/bin/phpunit:0
PHP   2. PHPUnit\TextUI\Command::main() /home/travis/.phpenv/versions/7.0.16/bin/phpunit:465
PHP   3. PHPUnit\TextUI\Command->run() phar:///home/travis/.phpenv/versions/7.0.16/bin/phpunit/phpunit/TextUI/Command.php:136
PHP   4. PHPUnit\Runner\BaseTestRunner->getTest() phar:///home/travis/.phpenv/versions/7.0.16/bin/phpunit/phpunit/TextUI/Command.php:159
PHP   5. PHPUnit\Framework\TestSuite->addTestFiles() phar:///home/travis/.phpenv/versions/7.0.16/bin/phpunit/phpunit/Runner/BaseTestRunner.php:65
PHP   6. PHPUnit\Framework\TestSuite->addTestFile() phar:///home/travis/.phpenv/versions/7.0.16/bin/phpunit/phpunit/Framework/TestSuite.php:400
PHP   7. PHPUnit\Util\Fileloader::checkAndLoad() phar:///home/travis/.phpenv/versions/7.0.16/bin/phpunit/phpunit/Framework/TestSuite.php:327
PHP   8. PHPUnit\Util\Fileloader::load() phar:///home/travis/.phpenv/versions/7.0.16/bin/phpunit/phpunit/Util/Fileloader.php:39
PHP   9. include_once() phar:///home/travis/.phpenv/versions/7.0.16/bin/phpunit/phpunit/Util/Fileloader.php:55
```

eperry:
I don't know how to do an update to a pull request,   so here is the info in words 

Had one small issue  in my ActiveDirectory we use sshPublicKeys  as the field hold multiple values and that is what the document we followed.   But in the code it calls it sshPublicKey

lib/functions.inc.php:    $userdata["sshPublicKey"] = $sshkey;

My change was 
lib/functions.inc.php:    $userdata["sshPublicKeys"] = $sshkey;

I suggest this might be a variable as I bet people call it other things too in their AD/Ldap

However, on my simple 2 test  it seems to work.  but have to do more testing but this is what I found so far

praymann:
> Had one small issue in my ActiveDirectory we use sshPublicKeys as the field hold multiple values and that is what the document we followed. But in the code it calls it sshPublicKey

Hrm, I'm against a 389-ds directory and was using the schema to line up with open-ldap:
```
dn: cn=schema
changetype: modify
add: attributetypes
attributetypes: ( 1.3.6.1.4.1.24552.500.1.1.1.13 NAME 'sshPublicKey' DESC 'MANDATORY: OpenSSH Public key' EQUALITY octetStringMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.40 )
-
add: objectclasses
objectClasses: ( 1.3.6.1.4.1.24552.500.1.1.2.0 NAME 'ldapPublicKey' SUP top AUXILIARY DESC 'MANDATORY: OpenSSH LPK objectclass' MUST ( uid ) MAY ( sshPublicKey ) )
```

Since it's doing a replace, see:
```
$userdata["sshPublicKeys"] = $sshkey;

$replace = ldap_mod_replace($ldap, $dn, $userdata);
```
Does it actually handle the attribute having multiple values? I expect that given the code in this PR, it will just replace the current value with the new value ( changetype: replace ). If the attribute has multiple values, it will delete them all and add the attribute back with a single new value.

To support multiple values, have to move from ldap_mod_replace to something like:
* grab all values of the attribute currently
* build array of the values, check array for new value
* add new value to array
* call ldap_mod_add with the array so the multiple values are applied

> I am not a PHP developer, so if this needs edits or changes let me know.

I can try to support a multi-value attribute, not sure how long it'll take me to codify and test it. :)

> I suggest this might be a variable as I bet people call it other things too in their AD/Ldap

Probably can just move into the config.inc.php file as a configuration option, with a sane default.

eperry:
We followed this guide, https://blog.laslabs.com/2016/08/storing-ssh-keys-in-active-directory/
for windows 2012,  I think the field while it might have a standard can be called anything.

> Does it actually handle the attribute having multiple values? I expect that given the code in this PR, it will 
> just replace the current value with the new value ( changetype: replace ). If the attribute has multiple 
> values, it will delete them all and add the attribute back with a single new value.

As for this, it will not replace an exsiting value, it returns a failed request. Tried that already :D this is also how ldapmodify works.   

In this case new attributes will just be appended in to the Field,  so a user can just keep adding new keys.

eperry:
*****Correction,  I just tested, the above "Removes the existing key" and puts the new key in place. It still rejects keys that are the same from being updated. I will see about testing what happens if the user already has Multiple Keys, will it remove all of them or just 1. I'd bet a cup of coffee that it removes them all.


I usually do a modify,  which it's behavior prevents the same key being uploaded ,and "Adds" new keys.So the account could have multiple sshkeys.

> changetype: modify
> add: sshPublicKeys
> sshPublicKeys: $key

I think replace should be used as normal because then we don't have to have a "Remove key" feature. Though "another variable would be good to flip this to modify  for people who want to have different keys for different pc's.

eperry:
Confirmed: the key removes all other keys and replaces with just the new one.

coudot:
I think we can just manage one key value for the first version of this feature

eperry:
Ok got one issue, I had an non-privileged user try to update their key and it failed with the following

> on Mar 13 11:37:20.116408 2017] [:error] [pid 3645] [client 10.233.138.248:50843] PHP Warning:  ldap_mod_replace(): Modify: Insufficient access in /var/www/reset.digital.hbc.com/lib/functions.inc.php on line 371, referer: https://reset.digital.hbc.com/?action=changesshkey
> [Mon Mar 13 11:37:20.116504 2017] [:error] [pid 3645] [client 10.233.138.248:50843] LDAP - Modify sshPublicKeys error 50 (Insufficient access), referer: https://reset.digital.hbc.com/?action=changesshkey
> 

I was going to look at the code, but seems like the app is not using the Admin Bind to do the key update? because I was able to replicate something similar with ldapmodify when binding as a normal user.

praymann:
> I was going to look at the code, but seems like the app is not using the Admin Bind to do the key update? 

Check what your "who_change_sshkey" property is set to in config.inc.php.

I mirrored the "who_change_password" property from the password change form.

In my environment, we defined an ACI to allow users to change their own sshPublicKey attribute, so the default is "user". The default for "who_change_password" is also "user".

eperry:
ah, got it that makes lots of sense will confirm tomorrow but we definately
dont do anything at the acl level so i must have it set to user.

i know only admin can do it in our domain. i might look to make the same
ACL allowance

menu.php
```diff
@@ -8,6 +8,7 @@
                 <span class="icon-bar"></span>
                 <span class="icon-bar"></span>
                 <span class="icon-bar"></span>
+                <span class="icon-bar"></span>
```
coudot: You don't need to add this span.

pages/changesshkey.php
```diff
@@ -0,0 +1,245 @@
...
+                <input type="password" name="password" id="password" class="form-control" placeholder="<?php echo $messages["password"]; ?>" />
+            </div>
+        </div>
+    </div>
+    <div class="form-group">
+        <label for="sshkey" class="col-sm-4 control-label"><?php echo $messages["sshkey"]; ?></label>
+        <div class="col-sm-8">
+            <div class="input-group">
+                <span class="input-group-addon"><i class="fa fa-fw fa-terminal"></i></span>
+                <input type="text" name="sshkey" id="sshkey" class="form-control" placeholder="<?php echo $messages["sshkey"]; ?>" />
```
coudot: We may need a textarea here

lib/functions.inc.php
```diff
@@ -358,12 +358,37 @@ function change_password( $ldap, $dn, $password, $ad_mode, $ad_options, $samba_m
...
 }
 
+
+# Change sshPublicKey attribute
+# @return result code
+function change_sshkey( $ldap, $dn, $sshkey ) {
+
+    $result = "";
+
+    $userdata["sshPublicKey"] = $sshkey;
```
coudot: We need a configuration parameter to set which attribute will hold the sshKey

coudot:
Hi @praymann,

sorry for the delay. This feature is very interesting and I would like to include it in 1.1 release. I made some comments in your code. Could you look at it and also try to fix conflicting files?

praymann:
I believe I have made the necessary changes per your review.

The conflicting files are fixed as well.

lib/functions.inc.php
```diff
@@ -363,6 +363,31 @@ function change_password( $ldap, $dn, $password, $ad_mode, $ad_options, $samba_m
     return $result;
 }
 
+
+# Change sshPublicKey attribute
+# @return result code
+function change_sshkey( $ldap, $dn, $sshkey ) {
```
coudot: We need to pass $change_sshkey_attribute as parameter of function change_sshkey, else it will be unef

pages/changesshkey.php
```diff
@@ -0,0 +1,245 @@
...
+            unset($extended_error);
+        }
+    }
+    if ( $errno ) {
+        $result = "badcredentials";
+        error_log("LDAP - Bind user error $errno  (".ldap_error($ldap).")");
+    } else {
+
+    # Rebind as Manager if needed
+    if ( $who_change_password == "manager" ) {
```
coudot: This should be $who_change_sshkey here.

coudot:
Thanks a lot for all your work!

Hope to release 1.1 soon.

plewin:
Hi @praymann 

Is it me or this link is corrupted for chinese translation ? https://github.com/ltb-project/self-service-password/pull/81/files#diff-266cb9bbae63984d8fdf4c86a9461091R122

https://github.com/ltb-project/self-service-password/pull/81/files#diff-9ce97ef165e01124464cb1f193dc6c01R382
If there is an error, does this log tells which attribute it tries to modify ? I think it could be misleading if the attribute is not sshPublicKey, maybe logging the attribute name would be a bit better ?

praymann:
> Is it me or this link is corrupted for chinese translation
Appears like you're correct, looks like URL encoding happened when I scripted up the translations.

> If there is an error, does this log tells which attribute it tries to modify ? I think it could be misleading if the attribute is not sshPublicKey, maybe logging the attribute name would be a bit better ?
Yes that could be misleading if someone changed the configuration to be another attribute.

@coudot I can fix these, I'll open a new PR.

PR #86, opened 2017-03-10
Enable credential-free deployment

LiptonB:
In my environment I would like to be able to deploy Self Service Password without storing any LDAP credentials on the box where it's running, and I realized that in many cases the user's own credentials should be enough. This modifies the tool so that `ldap_binddn` and `ldap_bindpw` can be left unset. Just thought I would share this in case you think it would be useful to others. Thanks for a great tool!

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -31,6 +31,8 @@ $ldap_base = "dc=example,dc=com";
 $ldap_login_attribute = "uid";
 $ldap_fullname_attribute = "cn";
 $ldap_filter = "(&(objectClass=person)($ldap_login_attribute={login}))";
+# DN that is the parent of all users
+$ldap_userparentdn = "cn=Users,dc=example,dc=com";
 
 # Active Directory mode
 # true: use unicodePwd as password field
--- a/pages/change.php
+++ b/pages/change.php
@@ -94,6 +94,9 @@ if ( $result === "" ) {
     # Bind
     if ( isset($ldap_binddn) && isset($ldap_bindpw) ) {
         $bind = ldap_bind($ldap, $ldap_binddn, $ldap_bindpw);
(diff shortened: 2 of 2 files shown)


coudot:
Hi,

why just don't allow anonymous search on LDAP directory to find user? Is this only a security issue?

LiptonB:
Yes, for security reasons we'd prefer not to have anonymous searches enabled. My thought was: if the user is already providing creds that can do the search, why open it up to everyone?

coudot:
We only have user credentials when user is changing password. For a reset, we don't have user password, so it won't work.

You can have a SSP account that has only search rights on directory.

LiptonB:
True, there's no way you would be able to do a reset without the app having its own credentials. We're getting around that by just having resets disabled, but I certainly acknowledge that resets are a nice feature that is missing when deployed this way.

An SSP account with only search rights would work for us, though it still wouldn't allow resets. I don't really see the advantage, though - fewer secrets makes it easier to deploy and the app still gets all the access it needs to do self-service changes. (Although I've just realized that a bad `$oldpassword` now gets you `ldaperror` instead of `badcredentials`, that might not be what we want...)

Anyway, no pressure. I realize the tradeoffs we've picked might not make sense for everyone. But if you like the idea and want any fixes to the code like the error message issue mentioned above, I'm happy to make them.

coudot:
I will see if we can include this in a future release, as this can only work with change password feature, and an LDAP directory with a single user branch

plewin:
Hi @LiptonB 

> Yes, for security reasons we'd prefer not to have anonymous searches enabled. My thought was: if the user is already providing creds that can do the search, why open it up to everyone?

What about allowing anonymous searches only from SSP's IP address ?

coudot:
We added a new option in #1129 to allow password change without any service account

Issue #85, opened 2017-03-09
mcrypt is outdated


dsclassen:
I've found many security people who recommend NOT using libmcrypt. It is aan old and unmaintained package.

https://paragonie.com/blog/2015/05/if-you-re-typing-word-mcrypt-into-your-code-you-re-doing-it-wrong

I am running php 7.0.10 on a CentOS7 machine and there is no mcrypt package available. I'm still trying to decide what to do, but in the meantime I thought I'd start a discussion here about potentially removing the mcrypt dependency for your project.

Cheers

coudot:
Indeed, we can use another library, we just need to rewrite encrypt/decrypt functions. Which other library is recommended?

plewin:
@coudot it depends on which php versions are supported.

no library is required for php >= 7.0
under 7.0, recommended extension is libsodium, it is not common. best alternatives are reading /dev/urandom or use mcrypt. openssl is not reliable on all php versions.

I tried to remove mcrypt and unsecure mt_rand for [#59](https://github.com/ltb-project/self-service-password/issues/59) unfortunately SSP use random bytes but also random integers, the code to implement them both ends up close to [random_compat](https://github.com/paragonie/random_compat)

I believe SSP should use random_compat to replace mt_rand and mcrypt. It will work out of the box for php 7 but for php < 7 it may require more work.

SSP can adopt 2 strategies for php < 7 :

- Request the installation of the mcrypt if php < 7.0 In order to be sure that a cryptography fonctions are present.
- Not installing anything and see if random_compat produces an exception at runtime. If an exception occurs, SSP would tell the user to install an extension, libsodium or simply mcrypt.

coudot:
Done by @plewin in PR #92

PR #92, opened 2017-03-21
Implements strong cryptography with defuse-crypto 2.0.3

plewin:
- Removes mcrypt and its mentions
- Uses random_int and random_bytes instead of mt_rand
- Update README
- Adds defuse-crypto + random_compat libraries

Uses defuse-crypto 2.0.3 (lib/defuse-crypto.phar), has random_compat embedded.

Solves :
https://github.com/ltb-project/self-service-password/issues/85 https://github.com/ltb-project/self-service-password/issues/60 https://github.com/ltb-project/self-service-password/issues/59

Impact for people already using SSP : reset tokens are now twice as large (75 -> 140 characters for me)

Impacts for new installations : mcrypt is no more required for encrypted tokens and sms feature.
Strong encryption functions are now required for random_compat (php 7 or libsodium or /dev/urandom readable or installing php-mcrypt)


coudot:
Hi @plewin , thanks a lot for this work.
Some remarks before merging:
* defuse is a third part library, it would be better to have it in lib/vendor instead of lib/
* We should write a simple unit test that call encrypt/decrypt methods and verify that we got the expected result (which is a string encrypted then decrypted is the same than the original)

plewin:
Hi @coudot, I updated the pull request with your remarks.

plewin:
@coudot please do not merge it now, i found something I want to improve

plewin:
I reproduced the old behavior, tempered encrypted tokens gives silent errors.
Initially in this pull request, exceptions were thrown. I prefer to no change the behavior in this PR.
This PR is ready for merge.

Issue #87, opened 2017-03-11
Get Travis tests working again on PHP 7


coudot:
We are not the only one to have the issue, see:
* https://core.trac.wordpress.org/ticket/40086
* https://core.trac.wordpress.org/ticket/39822
* https://github.com/sebastianbergmann/phpunit/blob/6.0/ChangeLog-6.0.md#600---2017-02-03

As we are not using composer for the moment, I see no simple solution.

I will remove for the moment tests on PHP 7 and let this issue open until we find a solution

coudot:
Done by @plewin in PR #99

PR #99, opened 2017-03-24
Force use of phpunit 5.7 if php >= 7.0 for travis testing

plewin:
Solves https://github.com/ltb-project/self-service-password/issues/87

Also add testing on php 7.1

Solution found in mantisbt
https://github.com/mantisbt/mantisbt/blob/master/.travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,13 @@ php:
   - '5.4'
   - '5.5'
   - '5.6'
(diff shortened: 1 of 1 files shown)


coudot:
Thanks. This indeed requires that Travis download phpunit for each test. The other solution would be to have phpunit in our lib/vendor.

PR #100, opened 2017-03-29
Fixes for things pointed out after #81 was merged

praymann:


praymann:
> It is still not valid html :/ This would be better

Forgive my mistake. Should be correct now per your suggestion.

plewin:
👍 and congratulations on making the pull request # 💯

Issue #57, opened 2016-12-06
Password History for SSHA


cybert0x:
Hello all,
It is not a issue but a feature which I miss in the project. 
In fact, I would like to check if the user use a password in History defined by pwdHistory in LDAP server. (about last 6 passwords for example)
I implement only for my need this patch with SSHA. Perhaps it could be an idea if you think the feature is useful for others persons, you can do a global patch for others encryptions.

Based on Self Service password 1.0 .

diff ./lib/functions.inc.php ./lib/functions.inc.php.ori
117c117
<     if ( preg_match( "/nophpldap|nophpmhash|ldaperror|nomatch|badcredentials|passworderror|passwordused|tooshort|toobig|minlower|minupper|mindigit|minspecial|forbiddenchars|sameasold|answermoderror|answernomatch|mailnomatch|tokennotsent|tokennotvalid|notcomplex|nophpmcrypt|smsnonumber|smscrypttokensrequired|nophpmbstring|smsnotsent|sameaslogin/" , $msg ) ) {
---
>     if ( preg_match( "/nophpldap|nophpmhash|ldaperror|nomatch|badcredentials|passworderror|tooshort|toobig|minlower|minupper|mindigit|minspecial|forbiddenchars|sameasold|answermoderror|answernomatch|mailnomatch|tokennotsent|tokennotvalid|notcomplex|nophpmcrypt|smsnonumber|smscrypttokensrequired|nophpmbstring|smsnotsent|sameaslogin/" , $msg ) ) {
237c237
< function change_password( $ldap, $dn, $password, $ad_mode, $ad_options, $samba_mode, $samba_options, $shadow_options, $hash, $hash_options, $who_change_password, $oldpassword, $login,$ldap_binddn, $ldap_bindpw) {
---
> function change_password( $ldap, $dn, $password, $ad_mode, $ad_options, $samba_mode, $samba_options, $shadow_options, $hash, $hash_options, $who_change_password, $oldpassword ) {
240,241c240
<     $password_used="";
<     $newpassword=$password;
---
>
333,348d331
<         $filter="uid=".$login;
<         $bind = ldap_bind($ldap, $ldap_binddn, $ldap_bindpw);
<         $search=@ldap_search($ldap, $dn, $filter, array ("pwdHistory") );
<         $entry = ldap_first_entry($ldap, $search);
<         $password_history = ldap_get_values($ldap, $entry, "pwdHistory");
<         foreach ($password_history as $pass) {
[...]

PR #101, opened 2017-03-30
Feature check password history

plewin:
PR for discussion. Not ready for merge.

Solves https://github.com/ltb-project/self-service-password/issues/57

What to do if there is a hash in password history uses a not supported scheme ? Ignore + log ?

Is the message "You already used this password"; sufficient ?

Is this implementation correct ? I tried to find inspiration in fusion directory but I fail to understand how  [they handle this](https://github.com/fusiondirectory/fusiondirectory/blob/df4f925d1c014d451844c646e70cdd7aa971f4e1/plugins/personal/generic/class_user.inc#L633). 

@cybert0x would you agree to test this branch ?

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -75,6 +75,11 @@ $hash = "clear";
 $hash_options['crypt_salt_prefix'] = "$6$";
 $hash_options['crypt_salt_length'] = "6";
 
+# Check password history, refuse password change if password is in password history
+$check_password_history = false;
+# Attribute with password history
+$password_history_attribute = "pwdHistory";
+
 # Local password policy
 # This is applied before directory password policy
 # Minimal length
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -76,9 +76,7 @@ $hash_options['crypt_salt_prefix'] = "$6$";
 $hash_options['crypt_salt_length'] = "6";
 
 # Check password history, refuse password change if password is in password history
-$check_password_history = false;
-# Attribute with password history
-$password_history_attribute = "pwdHistory";
+$ppolicy_options['check_password_history'] = false;
 
 # Local password policy
 # This is applied before directory password policy
--- a/lib/functions.inc.php
+++ b/lib/functions.inc.php
@@ -20,8 +20,8 @@
 #==============================================================================
 
 # Create SSHA password
-function make_ssha_password($password) {
-    $salt = random_bytes(4);
+function make_ssha_password($password, $salt = null) {
+    $salt = null === $salt ? random_bytes(4) : $salt ;
     $hash = "{SSHA}" . base64_encode(pack("H*", sha1($password . $salt)) . $salt);
     return $hash;
 }
@@ -39,8 +39,8 @@ function make_sha512_password($password) {
 }
 
 # Create SMD5 password
-function make_smd5_password($password) {
-    $salt = random_bytes(4);
+function make_smd5_password($password, $salt = null) {
+    $salt = null === $salt ? random_bytes(4) : $salt ;
     $hash = "{SMD5}" . base64_encode(pack("H*", md5($password . $salt)) . $salt);
     return $hash;
 }
@@ -52,7 +52,10 @@ function make_md5_password($password) {
 }
 
 # Create CRYPT password
-function make_crypt_password($password, $hash_options) {
+function make_crypt_password($password, $hash_options, $salt = null) {
+    if ( $salt != null ) {
+        return '{CRYPT}' . crypt($password, $salt);
+    }
 
     $salt_length = 2;
     if ( isset($hash_options['crypt_salt_length']) ) {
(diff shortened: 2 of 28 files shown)


conf/config.inc.php
```diff
@@ -75,6 +75,11 @@
 $hash_options['crypt_salt_prefix'] = "$6$";
 $hash_options['crypt_salt_length'] = "6";
 
+# Check password history, refuse password change if password is in password history
+$check_password_history = false;
+# Attribute with password history
+$password_history_attribute = "pwdHistory";
+
```
coudot: I would have done a $ldap_options[] array here, or ppolicy_options[].

As the code is done for the attribute pwdHistory syntax from standard LDAP password policy, I don't think there is a need to configure the name of the attribute.

We could just have for example:
````php
$ppolicy_options['check_password_history'] = true;
````

File: lib/functions.inc.php [hunk 2]
```diff
@@ -95,6 +98,93 @@ function make_ad_password($password) {
     return $adpassword;
 }
 
+# Returns an array with "scheme" and "salt"
+function user_password_analyzer( $user_password_value ) {
+    $matches = array ();
+    if ( !preg_match('/{(\S+)}(\S+)/', $user_password_value, $matches) ) {
+        throw new InvalidArgumentException('Hashed password does not validate LDAP format');
```
coudot: You throw an exception here but how do you catch it?

lib/functions.inc.php
```diff
@@ -95,6 +98,93 @@ function make_ad_password($password) {
...
+    $hash_details = user_password_analyzer($user_password_value);
+
+    $password = make_generic_password($cleartext, $hash_details['scheme'], $hash_details['salt']);
+
+    return $user_password_value == $password;
+}
+
+# Returns 'passwordused' if a hash for the $cleartext is present in $pwdHistory
+function check_password_history( $cleartext, $pwdHistory ) {
+    foreach ($pwdHistory as $oldPassword) {
```
coudot: You need to extract the password value from password history value, as it is done in FusionDirectory code.

Example of a value: `20170213101453Z#1.3.6.1.4.1.1466.115.121.1.40#4#test`

conf/config.inc.php
plewin: Okay 
I will keep `$ppolicy_options['check_password_history'] = true;`

lib/functions.inc.php
plewin: I do not plan to catch it, it is an assertion error. It should be catched by the exception_handler.
It will happen only the function is incorrectly used or the ldap server sends the pwdHistory in an incorrect format.

lib/functions.inc.php
plewin: Ok thank you, I understand better.
It works now because the preg_match skips the left part.
I will change the code to be more rigorous.

lib/functions.inc.php
coudot: How will this error be displayed in the interface? Will it be ignored and just logged?

coudot:
Hi @plewin, seems the unit test does pass with PHP 5.4, don't know why.

And you should also test passwordHistory with cleartext password. In this case there is no scheme, just the plain text value.

plewin:
Hi @coudot yes the PR is not ready yet.

It is failing because the function unpack have a different behavior since php 5.5. I will fix this.

plewin:
@coudot, I'm having doubts in the usefulness of this PR

I set up a Apache Directory Server for development, it is configured with ppolicy enabled by default and with pwdinhistory = 5.

Using the master branch,

If I try to change an user password with the previous password used, I get 
Red "Password was refused by the LDAP directory"

And it produces these lines in the logs

PHP Warning:  ldap_mod_replace(): Modify: Constraint violation in [..]\self-service-password\lib\functions.inc.php on line 448
[Sat Apr  1 18:10:51 2017] LDAP - Modify password error 19 (Constraint violation)

In this branch,

If I try to change an user password with the previous password used, I get 
Red "You already used this password"

(no error get logged because we don't try to modify).

I'm using who_change_password = user. When using who_change_password = manager, there is no constraint violation, any password can be set.

Maybe OP @cybert0x should try who_change_password = 'user' ?

coudot:
Hello @plewin 

Using the Directory Manager account will always bypass password policy. You need to either use the user account or a technical account which is not the Directory Manager.

You PR is very useful, because it allows to inform the user that the password is rejected because it is present in the history.

plewin:
Hi @coudot,

I tried again with a technical account, yes the ppolicy was not bypassed.

Would it be smart to, instead of this PR, display the technical messages LDAP_OPT_ERROR_STRING / LDAP_OPT_DIAGNOSTIC_MESSAGE under "Password was refused by LDAP directory" ? If the value of this PR  is only to show why the password was rejected, maybe it will be better ?

It could be enabled in config : $display_details_if_password_is_rejected_by_server = false...

Maybe it is possible to parse these messages to extract only the reason e.g. "invalid reuse of password present in password history". but it will require a vast amout of sample messages from different ldap vendors.

Some examples I got from rejected passwords on my Active Directory development server :
```
Password was refused by the LDAP directory
Details :
CONSTRAINT_VIOLATION: failed for MessageType : MODIFY_REQUEST Message ID : 5 Modify Request Object : 'uid=plewin,ou=People,dc=example,dc=com' Modification[0] Operation : replace Modification userPassword: 0x74 0x65 0x73 0x74 0x32 org.apache.directory.api.ldap.model.message.ModifyRequestImpl@2636fe49: invalid reuse of password present in password history
```

```
Password was refused by the LDAP directory
Details :
CONSTRAINT_VIOLATION: failed for MessageType : MODIFY_REQUEST Message ID : 5 Modify Request Object : 'uid=plewin,ou=People,dc=example,dc=com' Modification[0] Operation : replace Modification userPassword: 0x70 0x6C 0x6F 0x70 org.apache.directory.api.ldap.model.message.ModifyRequestImpl@2636fe49: Password should have a minimum of 5 characters
```

coudot:
Diagnostic message is enough with Apache DS but not with OpenLDAP. Each LDAP implementation will have a different message here. So if we really want to check password history, we need to manage this error ourself.

coudot:
Do you still want to work on this feature?

plewin:
@coudot Yes, I just got back from a short vacations and I need to install a new openldap for development

It looks like implementation of pwdHistory differs in some LDAP servers

Apache Directory Server encodes the userPassword field in base64
https://github.com/apache/directory-server/blob/7391447b3cee132bbcc9a023f435919025c28511/interceptors/authn/src/main/java/org/apache/directory/server/core/authn/PasswordHistory.java#L75

Should this feature be supported for only OpenLDAP ?

coudot:
We should rely on ppolicy draft : https://tools.ietf.org/html/draft-behera-ldap-password-policy-10

> 5.3.5.  pwdHistory
> 
>    This attribute holds a history of previously used passwords.  Values
>    of this attribute are transmitted in string format as given by the
>    following ABNF:
> 
>       pwdHistory = time "#" syntaxOID "#" length "#" data
> 
>       time       = GeneralizedTime
> 
>       syntaxOID  = numericoid    ; the string representation of the
>                                  ; dotted-decimal OID that defines the
>                                  ; syntax used to store the password.
> 
>       length     = number        ; the number of octets in data.
> 
>       data       = <octets representing the password in the format
>                     specified by syntaxOID>.

Maybe ApacheDS is not compliant with it, and we need in this case an option to adapt our code to ApacheDS.

coudot:
Hello @plewin, do you think you can resolve conflicts?

plewin:
I am still working on this feature.

These are the 2 points why I'm not okay yet.

- openldap vs apache directory implementation difference, I am still undecided on what is the best move to support both. apache directory has a extra base64encode in their passwords. I am still looking for a magic and correct way to guess without performance penalties and hacks. Otherwise, I will have to add an extra configuration variable but I'm unsure what I should ask. $is_apache_directory à la $is_ad or something like $ldap_server = 'openldap'
- I would like to test on a fusiondirectory's openldap, fusiondirectory hacks the userpassword attribute to mark an user locked. I want to be sure that this feature works well with an openldap managed by fusiondirectory.

coudot:
>  openldap vs apache directory implementation difference, I am still undecided on what is the best move to support both. apache directory has a extra base64encode in their passwords. I am still looking for a magic and correct way to guess without performance penalties and hacks. Otherwise, I will have to add an extra configuration variable but I'm unsure what I should ask. $is_apache_directory à la $is_ad or something like $ldap_server = 'openldap'

As OpenLDAP respect the ppolicy draft, I would say that SSP code should by default be comaptible with OpenLDAP. We can indeed add an option for ApacheDS to support their own history syntax.

> I would like to test on a fusiondirectory's openldap, fusiondirectory hacks the userpassword attribute to mark an user locked. I want to be sure that this feature works well with an openldap managed by fusiondirectory.

Using ppolicy, fusiondirectory should instead use the pwdAccountLockedTime to lock the account. Are you sure it changes the userPassword value?

plewin:
> Using ppolicy, fusiondirectory should instead use the pwdAccountLockedTime to lock the account. Are you sure it changes the userPassword value?

You made me doubt but no it really works like that in fusiondirectory, probably because locking was supported before ppolicy.

It makes me realize that one can unlock himself from a fusiondirectory managed openldap using SSP and the reset by SMS feature. Perhaps this should be mentionned in the doc or the reset be prevented if the user is locked in fusiondirectory, in an other PR.

```
/*! \brief       Locks an account by adding a '!' as prefix to the password hashes.
 *               This makes login impossible, due to the fact that the hash becomes invalid.
 *               userPassword: {SHA}!q02NKl9IChNwZEAJxzRdmB6E
 *               sambaLMPassword: !EBD223B61F8C259AD3B435B51404EE
 *               sambaNTPassword: !98BB35737013AAF181D0FE9FDA09E
```
https://github.com/fusiondirectory/fusiondirectory/blob/master/include/password-methods/class_password-methods.inc#L93

coudot:
> It makes me realize that one can unlock himself from a fusiondirectory managed openldap using SSP and the reset by SMS feature. Perhaps this should be mentionned in the doc or the reset be prevented if the user is locked in fusiondirectory, in an other PR.

Indeed, but locking feature from FusionDirectory is not very standard. I don't know if we need to mention it in SSP documentation. And FusionDirectory has its own password recovery self service.

bilbo-the-hobbit:
Hello,

if fusiondirectory does in fact touch the userpassword when ppolicy plugin is actived and the overlay inside openldap is used, there is a bug around that.

of course Fusiondirectory not only lock only userpassword but the samba account and ssh key

feel free to enter a bug at https://forge.fusiondirectory.org/

Cheers

plewin:
Closing this PR because a better alternative is now available.

Users wanting to check the password history should use ppolicy.

`$ldap_use_ppolicy_control = true;`

PR #102, opened 2017-04-10
Fix for base64 encoded strings that contain '+'

Koshatul:
When the base64_encoded string contains +, when it's sent to be decoded it comes back with "ciphertext too short"

This may be an issue with PHP 7.0+, but in theory it should be a problem for 5.6 as well.


coudot:
Thanks for the patch, could you also apply it on sendsms.php?

Koshatul:
Fixed sendsms.php as well, waiting on travis.

coudot:
Thanks for the patch!

PR #104, opened 2017-04-11
Fix invalid html in sendsms.php

plewin:
Related to https://github.com/ltb-project/self-service-password/pull/102

I am not sure if there is a bug, the html could be misinterpreted if the base64 contains "/"


coudot:
Yes it's better now

PR #107, opened 2017-05-03
Sanitize Mobile Number retrieved from LDAP

kidmock:
I have the issue that mobile numbers stored in LDAP are variable format as hard as I have tried to get people. The format is always different.
1(NNN) NNN-NNNN
+1 NNN.NNN.NNNN
NNN-NNN-NNNN

etc.

This simple patch sanitize the number format so email to sms will work no matter how it's stored. 
It can then be converted to 1NNNNNNNNN@smsemaildomain.foo without needing constantly 'fix' the mobile attribute in LDAP.

Cheers

### DIFF
--- a/pages/sendsms.php
+++ b/pages/sendsms.php
@@ -180,6 +180,8 @@ if ( $result === "" ) {
     # Check sms number
     if ( $smsValues["count"] > 0 ) {
         $sms = $smsValues[0];
+          $sms = preg_replace('/[^0-9]/','',$sms);
+          $sms = substr($sms, -10);
     }
 
     if ( !$sms ) {


coudot:
The feature is interesting, but your code fix the number to 10 digits, which is maybe not the case for everyone.

I think we can create a configuration option to enable this feature and maybe also an option to fix the number size.

kidmock:
I figured the U.S. 10 digits limit might be an exception. I thought I'd at least shared the patch I apply after every update. Maybe others will find it useful. Or perhaps, as you suggested, it can be a configurable item.

Unfortunately, I'm not sure how email to sms is structured outside of the U.S. I would think non digits should be excluded?

Cheers

coudot:
I added configuration parameters and merged it.

Thanks.

PR #108, opened 2017-05-06
SSH key validation

Kage:
I updated the SSH Key page to support key bit size checking, as well as some updates for better wording.

I have not yet done the additional languages.  I can do that in a separate pull request if desired, though be warned it will be through Google Translate. ;)

Edit: Looks like I have to do translations. @_@ Ok, one moment...


coudot:
Thanks a lot, I will look at it.

Kage:
Just pushed an update, because somehow I forgot functions.inc.php :/

lib/functions.inc.php
```diff
@@ -239,6 +260,29 @@ function check_password_strength( $password, $oldpassword, $pwd_policy_config, $
...
 }
 
+function check_sshkey( $ssh_pubkey, $sshkey_policy_config ) {
+    extract( $sshkey_policy_config );
+
+    # Split up the SSH key
+    $parts = explode(' ', $ssh_pubkey);
+
+    # Check this gave us at least two parts, and SSH key type is valid
+    if (count($parts) < 2 || $parts[0] != 'ssh-rsa') { return "keyinvalid"; }
```
plewin: The function considers all non rsa keys as invalid. I do not think it is acceptable to disallow other types of keys : dsa, ecdsa and ed25519.

plewin (changes requested):
Other types of keys need to be considered before merging this. Also unit testing of the function that validates the keys would be greatly appreciated.

coudot:
@Kage and @plewin, do you think you will have time to work on this feature?

Kage:
Completely forgot about this, I'll look into it soon. :P

Issue #115, opened 2017-06-08
Force specific language?


sfieux:
I'm trying to force the use of english for my installation of self-service-password. It seems that this piece of code from `index.php` is causing my trouble:

```
require_once("conf/config.inc.php"); # contains line $lang ="en"; 
(...)
$lang = detectLanguage($lang, $languages);
require_once("lang/$lang.inc.php");
```

The way I understand it is: the language is defined in `config.inc.php`, and afterwards this information is discarded to read it from browser settings in `detectLanguage`.

Am I missing something?

If I'm not, would you be interested in a PR to change this behaviour? I would change the default language value to "auto" in `config.inc.php` and only try to detect the browser language if the value is different:

```
if ( $lang == "auto" ) {
    $lang = detectLanguage($lang, $languages);
}
require_once("lang/$lang.inc.php");
if (file_exists("conf/$lang.inc.php")) {
    require_once("conf/$lang.inc.php");
}
```

coudot:
Maybe the best would be to configure a list of authorized languages. Then this would require a new parameter which will be an array. If it is empty, all available languages will be authorized, else this parameter will be used to limit which languages could be used.

sfieux:
What would be the expected behaviour when several languages are authorized but the detected one is not part of the list? Une the first authorized one? Use initial `$lang` value?

coudot:
if the lang is not one authorized, then fall back to default lang

plewin:
I'm closing this issue because the PR was merged.
Thank you for your contribution @sfieux.

sfieux:
No problem, I'm the one who needed it :)

PR #116, opened 2017-06-08
Add possibility to force use of a specific set of languages

sfieux:
As disscussed in issue #115.

Tested following cases:
* One of the browser languages is in the allowed languages => browser language selected;
* None is => default language is used;
* Default language is/is not included in the allowed languages => in both cases, browser or default language is selected, depending if browser language is in allowed languages list ;
* Default language matches / does not match an existing lang file => in the unlikely case where the browser language doesn't match any allowed languages, and the default languages does not have any `$lang.inc.php` file, fall back to english

Any comments are welcome

### DIFF
--- a/index.php
+++ b/index.php
@@ -38,16 +38,24 @@ require_once("lib/vendor/PHPMailer/PHPMailerAutoload.php");
 $languages = array();
 if ($handle = opendir('lang')) {
     while (false !== ($entry = readdir($handle))) {
-        if ($entry != "." && $entry != "..") {
-             array_push($languages, str_replace(".inc.php", "", $entry));
+        if ($entry != "." && $entry != ".." ) {
+            $entry_lang = str_replace(".inc.php", "", $entry);
+            # Only add language to possibilities if it is the default language or part of the allowed languages 
+            # empty $allowed_lang <=> all languages are allowed
+            if ($entry_lang == $lang || empty($allowed_lang) || in_array($entry_lang, $allowed_lang) ) {
+                array_push($languages, $entry_lang);
+            }
         }
     }
     closedir($handle);
 }
-if(!isset($lang) or $lang == "auto") {
-    $lang = detectLanguage($lang, $languages);
+$lang = detectLanguage($lang, $languages);
+if (file_exists("lang/$lang.inc.php")) {
+    require_once("lang/$lang.inc.php");
+} else {
+    # Fall back to english if default $lang does not actually exist
+    require_once("lang/en.inc.php");
 }
-require_once("lang/$lang.inc.php");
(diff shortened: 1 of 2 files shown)


index.php
```diff
@@ -38,14 +38,24 @@
...
+            if ($entry_lang == $lang || empty($allowed_lang) || in_array($entry_lang, $allowed_lang) ) {
+                array_push($languages, $entry_lang);
+            }
         }
     }
     closedir($handle);
 }
 $lang = detectLanguage($lang, $languages);
-require_once("lang/$lang.inc.php");
+if (file_exists("lang/$lang.inc.php")) {
```
plewin: I do not see a value in this if / else block.

$lang here is either the default $lang from the config file or one from the $languages array.
All files with $lang from $languages exist because they are only added to in the while loop reading the files inside the _./lang_ dir.

So the only way for this expression to return false is that the default $lang from config file has no associated lang/...inc.php file. That's probably a typo e.g. `$lang = "En";` or the associated language file does not exist.
In all cases, there is a configuration error. I would rather log this e.g "please fix $lang from config" or leave the `require_once("lang/$lang.inc.php");` so PHP will halt the script and tell sorry there is a missing file.

sfieux: I don't like the idea of leaving the `require_once("lang/$lang.inc.php");` alone. This would leave a blank page, without any message at all for the site admin to investigate what's happening.

I would rather add a line to the error log when switching to english. This would be more helpful to investigate why the default language is not selected.

plewin: > This would leave a blank page, without any message at all for the site admin to investigate what's happening.
Actually it depends on the php configuration. I just realized that the error reporting is configured after and not before, it should be fixed before the language section. I expected the require_once error to be silent in production and displayed and logged when $debug = true.

I don't believe SSP should try to fall back to english when it knows that a configuration/lang file is incorrect or incomplete. But that's my point of view. What do you think @coudot ?

I have a new suggestion, instead of checking if the file exists, before detectLanguage, why not check if $lang is in $languages. If it is not, the language file does not exists so SSP can either : 1) die early and tell there is a configuration error that should be fixed 2) do what you prefer, fall back to english and log

The problem with the current code is that in some edge use cases ex.
Administrator is using english. Most users are japanese. Default language is set to jap instead of ja.
The administrator will notice that the default language is not working on later in production logs when the first non english/japanese use SSP.

coudot: @plewin you are right for error reporting, it should happen before the lang selection.

And yes with error reporting, we will know if the lang configuration is good or not. The failback to the default lang is already coded in the detectLanguage, I agree we should not force a failback on English after.

sfieux:
As discussed, I moved the error reporting before the language selection and removed the "fall back to english" part.

PR #134, opened 2017-08-10
add ldap extended error

konvergence:
Hi,

I use OpenLDAP with ppolicy.
When use self-password , and get error from openldap, cause of "password policy" into openldap, "self-password return only "Password was refused by the LDAP directory".

Could you get the extended error to be able to display the message coming from openldap ?

I do the following change into "change.php", but I suppose that I must do also "setquestions.php", "resetbytoken.php" and "resetbyquestions.php" ?

### DIFF
--- a/pages/change.php
+++ b/pages/change.php
@@ -19,6 +19,11 @@
 #
 #==============================================================================
 
+# define not available in php 5
+define("LDAP_OPT_DIAGNOSTIC_MESSAGE", 0x0032);
+
+
+
 # This page is called to change password
 
 #==============================================================================
@@ -190,6 +195,12 @@ if ( $result === "" ) {
         $command = escapeshellcmd($posthook).' '.escapeshellarg($login).' '.escapeshellarg($newpassword).' '.escapeshellarg($oldpassword);
         exec($command);
     }
+ 
+    $ldap_extended_error = "" ;
+    if ( $result === "passworderror" ) {
+      ldap_get_option($ldap, LDAP_OPT_DIAGNOSTIC_MESSAGE, $ldap_extended_error);
+   }
+
 }
 
 #==============================================================================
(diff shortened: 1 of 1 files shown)


konvergence:
Hi all,

On my side , I get several "Password was refused by the LDAP directory" and , not only in case of password history. Example : 
- [Password fails quality checking policy]
- [Password is in history of old passwords]
- [Password is too young to change]

But I can said if there is no potential errors

pages/change.php
```diff
@@ -19,6 +19,11 @@
 #
 #==============================================================================
 
+# missed defines in php 5
+define("LDAP_OPT_DIAGNOSTIC_MESSAGE", 0x0032);
+
```
coudot: Would have prefer this:
````php
if ( !defined("LDAP_OPT_DIAGNOSTIC_MESSAGE") ) {
    define("LDAP_OPT_DIAGNOSTIC_MESSAGE", 0x0032);
}
````

And this code should be in index.php to avoid duplication in several files.

File: pages/change.php [hunk 2]
```diff
@@ -198,7 +210,7 @@
 ?>
 
 <div class="result alert alert-<?php echo get_criticity($result) ?>">
-<p><i class="fa fa-fw <?php echo get_fa_class($result) ?>" aria-hidden="true"></i> <?php echo $messages[$result]; ?></p>
+<p><i class="fa fa-fw <?php echo get_fa_class($result) ?>" aria-hidden="true"></i> <?php echo $messages[$result] . " " . $ldap_extended_error ;  ?></p>
```
coudot: As previously said in comment, the extended error should not be shown to end user, but can be logged for administrators

But if you can parse the extended message, you may assign a different error message (`tooshort`, `minlower`, etc.) instead of the default one (`passworderror`).

pages/change.php
konvergence: May be it will better to define in lib/function.inc.php ?

pages/change.php
konvergence: You are right. It is better to interpret the text returned by OpenLDAP, in order to return a translated text. But I would post the original message if I do not have a match.

pages/change.php
plewin: yes put it functions.inc.php please, so unit tests can use this constant too

konvergence: Done

coudot:
Was done in #367

PR #138, opened 2017-08-17
sendtoken.php send http instead of https

konvergence:
Please find a proposal to handle HTTP_X_FORWARDED_PROTO and HTTP_X_FORWARDED_PORT
I know that I can also use the $reset_url also

### DIFF
--- a/pages/sendtoken.php
+++ b/pages/sendtoken.php
@@ -182,14 +182,21 @@ if ( $result === "" ) {
 
     if ( empty($reset_url) ) {
 
+        $server_name = $_SERVER['SERVER_NAME'];
+        $server_port = $_SERVER['SERVER_PORT'];
+        $script_name = $_SERVER['SCRIPT_NAME'];
+        
         # Build reset by token URL
         $method = "http";
         if( !empty($_SERVER['HTTPS']) || (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https')){
            $method .= "s";
         }
-        $server_name = $_SERVER['SERVER_NAME'];
-        $server_port = $_SERVER['SERVER_PORT'];
-        $script_name = $_SERVER['SCRIPT_NAME'];
+
+
+        # change servername if HTTP_X_FORWARDED_HOST is set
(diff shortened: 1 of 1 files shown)


coudot:
@plewin what do you think this? Should we test be default the X_FORWARDED headers to avoid to configure the reset_url? The reset_url will then be only useful for very specific needs.

plewin:
@coudot @konvergence 
I believe that support of reverse proxy to detect the reset url is interesting and useful in environments when less configuration is more and proxies are omnipresents such as docker environments.

When reading the code, I really do not feel safe when I read `$_SERVER['HTTP_X_FORWARDED_HOST/PORT']`. They are almost literally user inputs and should not be processed without extreme care and without knowledge of the architecture where SSP is running on. 

If this PR get merged as it, it will be possible to send to any user you know their login + password a link to a phishing page by manipulating your http headers -- assuming the web server does not strip them before submitting the request to php. That is absolutely not secure by default.

Here some exemples how big projects deal with this :
Zend Framework do not process them unless you specify an useProxy = true which is false by default.
Symfony do not process them unless you provide a whitelist of proxy servers that are allowed to set this variables.
Drupal 8 configuration file has a long and descriptive comment/explanation about this issue. 
https://github.com/drupal/drupal/blob/4a629e32dfe0d29ff14c756e65f196d5b9947036/sites/default/default.settings.php#L343

I suggest that either :
Update the PR to process these variables only if ssp is told (configured) so.
Close this PR because the usefulness is limited.

In my opinion, this PR can be closed.

konvergence:
@plewin

You  should also update the documentation,  because  the current suggest to  build reset_url with HTTP_X_FORWARDED_PROTO/HOST/PORT without any explanation of the risks.

You're right about docker , I currently test SSP on it.
My need was to build the reset_url but only with HTTP_X_FORWARDED_PROTO and HTTP_X_FORWARDED_PORT. Because HTTP_X_FORWARDED_HOST was not set.

I will add an option into config.inc.php to enable the code and see if i can manage a whitelist on remoteip

coudot:
Hello @konvergence, we think this PR is not very useful as we already have the `$reset_url`. We do not want to add to much complexity in configuration (adding proxy option and whitelist option). Thanks anyway for the work and the interesting discussion on this topic.

PR #139, opened 2017-08-21
check password expiration

konvergence:
Hi,

This is a proposal to replace the shell script.
The  schedule can be done with a curl in POST to provide a login/password member of an admin group.

This can be also be adapt to allow ad_mode. But I did'nt work it yet.
There is also some message to create.


coudot:
This will be done in #1047 with a script on server side

Issue #128, opened 2017-07-25
Reset token validation issue


lucamaro:
Hi all,
I have trouble in reset password via email, basically the token is never considered valid because user session variables (login and time) are not saved to disk (session file is empty). Enabling error_reporting with E_ALL the page says:

`Warning: session_start(): Cannot send session cache limiter - headers already sent (output started at /var/www/html/index.php:155) in /var/www/html/pages/sendtoken.php on line 149`

Is there a problem or I'm missing something? BTW I tried ssp v1.0, but it seems to me that master branch has the same behaviour.

Thank you for this great work.
Luca

lucamaro:
So, after some work I confirm that the problem is due to session_start in sendtoken.php called after the beginning of html page, as the error message asserts.

With some changes I got it working, if you want I can submit a PR. But some of the logic about token/session inside sendtoken.php has to be moved into index.php.

Let me know.
Luca

plewin:
Hi @lucamaro, thank you for posting this.

I acknowledge that the code dealing with the sessions/tokens has some problems like being difficult to understand and producing unnecessary alarming logs but it should work.

Would you mind posting a link to the commit/branch working for you and posting your phpinfo details on a github gist ?

lucamaro:
Hi @plewin,
here is [phpinfo](https://gist.github.com/lucamaro/459585a53522a2cb49d4d9d5278f1939) and here a super minimal change to [index.php](https://github.com/lucamaro/self-service-password/commit/2065f6f5d1b7fe4686bef8f316c2b2cb68637874) that's working for me.

Without this change session files are always empty, so no session data is saved and reset token page says that the token is not valid.

Thank you!
Luca

coudot:
I don't see why the current code fails. We use the ini_set calls just before calling cookies method. Maybe there are other cookies used in the application (set by a reverse proxy in front of SSP) which could explain why it fails?

lucamaro:
I'm using docker with a custom [lucamaro/ltb-self-service-password](https://hub.docker.com/r/lucamaro/ltb-self-service-password/) based on standard php:7-apache, no reverse proxy in front of it.

I tried [moss/selfservicepassword](https://hub.docker.com/r/moss/selfservicepassword/) that is based on php:7-fpm and nginx as reverse proxy and it does not have the same problem. 

Which setup do you suggest? 

BTW, start_session documentation has the following note:

>     Note:
[...]

PR #146, opened 2017-08-31
Output buffering to avoid failing session_start in PHP 7.1

plewin:
Implements ob_start / ob_end_flush as discussed in https://github.com/ltb-project/self-service-password/issues/128

Revert "output_buffering" for php.ini because it is no longer needed.

Also remove redundant closing php tags to avoid outputing unintentionally white spaces


plewin:
It is a good practice (i agree, sometimes questionned), actually it is not necessary because I added the ob_start before the includes().

There is no problem if there is close tag after ob_end_flush() but it is the only file I'm okay with. Other files are better without the risk of unwanted whitespaces that may cause problems in future unit tests.

`If a file is pure PHP code, it is preferable to omit the PHP closing tag at the end of the file. This prevents accidental whitespace or new lines being added after the PHP closing tag, which may cause unwanted effects because PHP will start output buffering when there is no intention from the programmer to send any output at that point in the script.`
https://secure.php.net/manual/en/language.basic-syntax.phptags.php

coudot:
Ok I did not know that!

I will merge this PR and close all issues related to 1.1. We can know release this version.

Thanks a lot for this work!

PR #149, opened 2017-09-01
Remove obsolete stripslashes_if_gpc_magic_quotes

plewin:
Please remove magic quotes support, it is obsolete/removed since php 5.4.

If some users want to use the code en PHP 5.3, which I do not recommend because we do not test + this php version is no more secure + ssp only work partially on this version. Let them set magic quotes = off in their php.ini.

Keeping these functions is harmful to SSP's future

### DIFF
--- a/lib/functions.inc.php
+++ b/lib/functions.inc.php
@@ -141,16 +141,6 @@ function generate_sms_token( $sms_token_length ) {
     return $smstoken;
 }
 
-# Strip slashes added by PHP
-# Only if magic_quote_gpc is not set to off in php.ini
-function stripslashes_if_gpc_magic_quotes( $string ) {
-    if(get_magic_quotes_gpc()) {
-        return stripslashes($string);
-    } else {
-        return $string;
-    }
-}
-
 # Get message criticity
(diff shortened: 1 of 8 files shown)


plewin:
Last time I checked for LDAP injections I found none.

The is a ldap_espace in php 5.6 which can be useful
https://secure.php.net/manual/fr/function.ldap-escape.php
Symfony provides a polyfill to support previous php versions. It comes from this stackoverflow post https://stackoverflow.com/questions/8560874/php-ldap-add-function-to-escape-ldap-special-characters-in-dn-syntax/8561604#8561604

coudot:
Indeed we have no LDAP injection as we check some forbidden chars in login.

ldap_escape can be a good thing to support logins like domain\user, which for the moment will create this error:

> [Mon Sep 04 11:11:15.480641 2017] [:error] [pid 4923] [client 127.0.0.1:39030] PHP Warning:  ldap_search(): Search: Bad search filter in /home/clement/dev/self-service-password/pages/change.php on line 99, referer: http://ssp.example.com/

We can use it if it exists (and do not escape values for old PHP versions).

PR #150, opened 2017-09-02
Extract templates from pages and switch to Twig

plewin:
This PR is a lot of refactor + switching to twig for templates.

I added Twig in a phar in lib.

I left some NOTICES if $debug = true, I will fix them in an other PR after this one.

Performance is not optimal because I do not use caching yet and that is also for an other PR.


plewin:
All "Expected next thing to be an escaping function" from codacy are false positives.
It works the other way in Twig. All outputs are secure by default. One must use the filter "|raw" to ask twig to not escape.

Issue #155, opened 2017-10-23
Use password modify extended operation


coudot:
In PHP 7.2 we will be able to use password modify extended operation. We need to add an option to be able to use it (and test if the function is available).

plewin:
I tried the functions on a php 7.2.1 / apache directory server 2.0.0-M24 on windows

ldap_exop_password did not work because of some "Decoding error"
https://gist.github.com/plewin/3ecfcacb2d7846d28598859f90919898#file-trace-log-L54

ldap_exop with LDAP_EXOP_MODIFY_PASSWD worked but when the password cannot be modified because of password policy, ldap_exop said it was still a success. It is not possible to know in php if the password was really changed or not, nor getting the details of the problem without using ldap_get_option with LDAP_OPT_DIAGNOSTIC_MESSAGE
https://gist.github.com/plewin/3ecfcacb2d7846d28598859f90919898#file-trace-log-L91
https://gist.github.com/plewin/3ecfcacb2d7846d28598859f90919898#file-trace-log-L98

bilbo-the-hobbit:
hello,

if you have issues with php 7.x an php ldap you can ask the developper directly @MCMic

Cheers

MCMic:
@plewin Hello, can you please open tickets on https://bugs.php.net/ about these problems?
EXOP support code is quite new and might indeed need a few fixes in the PHP versions to come.

PR #157, opened 2017-10-24
Using ldap_exop_passwd if available (PHP>=7.2)

MCMic:
Attempt at #155 
This is not tested.

### DIFF
--- a/lib/functions.inc.php
+++ b/lib/functions.inc.php
@@ -265,7 +265,7 @@ function change_password( $ldap, $dn, $password, $ad_mode, $ad_options, $samba_m
             if ( isset($userpassword) ) {
                 if ( preg_match( '/^\{(\w+)\}/', $userpassword[0], $matches ) ) {
                     $hash = strtoupper($matches[1]);
-		}
+                }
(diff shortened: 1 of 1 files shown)


coudot:
Thanks but we need a more complex code: even if the function exists, we want to be able to use it per configuration settings, as we might have an LDAP server that does not support the exop.

Anyway, we can use your code as implementation example when we will work on the feature.

dgeo:
Would you prefer a "exop" value for $hash (as using exop removes the need for hashing passwords) ?
I may PR this quickly…

dgeo:
Another tentative implementation here: https://github.com/criecm/self-service-password/commit/bda96b1a917b4ab475f93ccc8317cfd9662ba707
The idea is tu use 'exop' as hash value to use it

coudot:
My idea is to add an option to use the exop method, it is not linked to the hash method.

dgeo:
it may be viewed as 'the server will hash it himself', we need to keep it clear until that ;)
I wrote this one quickly because I needed it…

PR #158, opened 2017-10-25
Controllers + Dependency Injection with Pimple

plewin:
PR for discussion

These changes are a new step towards more modern code with Silex (or not, or another)

The SSP pages were split php / html -> in a previous PR introducing twig https://github.com/ltb-project/self-service-password/pull/150, in this PR, the remaining pages were converted from procedural code to controllers classes.

To make the controllers classes Symfony/Silex style/compliant. The classes Request/ParameterBag/Controller were created in index.php. They are super minimal classes that mimics symfony http api. These classes can be moved to their own file later e.g. Framework.php or be removed if Silex or similar is adopted.

To keep the controllers as short as possible in the idea of "thin controllers, fat model" most of the logic inside the controllers were extracted to service classes. I put these classes in functions.inc.php but they will need a new file for each.

I believe there are new regressions I still haven't catched but I am not worried because testability is almost achieved.

With new controllers classes and service classes containing the previous logic in the controllers, dependency injection with pimple (Silex's Dependency Injection Container) was introduced. The new DIC allows to decouple the code and simplify configuration of services, unit testability and override of parts of SSP (which can be really useful for testing or to host a demo version with a fake ldap/emails).
For now, Pimple is shipped with a phar, similarly to Twig & Defuse-Crypto, composer makes sense but I don't want to touch how ssp is packaged for now.

Now I would like to make a change in the file structure. I would like SSP to use namespaces and use them with an autoloader instead of require(). \App or \LdapToolBox or \SelfServicePassword
The structure would look like
```
./conf // no changes
./css // i would like them in a ./web but not for now because it changes the packaging
./fonts // i would like them in a ./web but not for now because it changes the packaging
./images // i would like them in a ./web but not for now because it changes the packaging
./js // i would like them in a ./web but not for now because it changes the packaging
./lang
./lib // ideally remove it and move everything inside in src
./lib/vendor // i would like it to be moved to /vendor but I'm not sure if it changes the packaging
./packaging
./pages // moved into src
./src // new folder with all the ssp code
./src/Controller // new home for the controllers
./src/Framework // Symfony/Silex compatibility classes
./src/Service // services classes used by controllers
./templates
./tests
./var
```

@coudot Is this the right direction for SSP ?


coudot:
Hello @plewin, thanks for all this work.

It looks good to me, and we can easily adapt packaging to new structure. I am not an expert in PHP development, so I don't have any good feedback about pimple and other coding choice. But I agree we need to really split code from templates and have better test coverage, so your proposal is very interesting.

I think we can plan it for 1.2 if you think you can work on this during next weeks.

plewin:
I made my suggested new code structure in the PR. I also added couple new tests to illustrate how unit testing can be done with the new structure.

huh are you planning to release 1.2 already ?

coudot:
There is no hurry to release 1.2. But if you want to change the code organization, we should do it quick to allow Pull Requests on this new code. Else you will need to merge all changes inside you own Pull Request.

plewin:
This PR is ready for merge in my opinion

I'm thinking about working on these subjects next in approximately this order :
* replace error_log by PSR-3 compliant logger
* enforce coding standards and style from PSR 1 & PSR 2 + adopt short array notation array() -> []
* move token related code to their own classes, because they use globals, init_set which still cause problems for testability, some controllers are not unit testable because of that
* move templates to templates/ssp-default or similar and move css,fonts,images,js to templates/ssp-default. Allow selection of theme in config.
* make use of exceptions instead of string error codes
* more more unit tests
* acceptance test, with a fake browser testing the app with specific scenarios.

coudot:
Hello @plewin,

to be honest, I will have no time to review what you have done, but it seems to be a good move and I think you can merge it.

plewin:
Closing this for now, I will submit these commits to a 2.0 branch

PR #162, opened 2017-11-14
Resolve send token web page issue when E-Mail To: set from LDAP

Jitsusama:
There is an issue where the info message on the `sendtoken.twig` generated webpage triggered by clicking on the *Email* link on the main page shows the message **"Enter your user name and your email address to reset your password..."**, even when the `$mail_address_use_ldap` `config.inc.php` parameter is set to `true`. In this case, the email address is automatically deduced from LDAP and is not presented to the user to change, so the text message should not state otherwise.

This PR adds a new translatable message keyed with the string `'sendtokenhelpnomail'`. This message *should* explain that only a username needs to be supplied, while also hinting that the user's email address will be automatically determined by some other means.

I have added the translation for `lang/en.inc.php` under the proper key. For all other languages, I made the new key, but copied the value from the `'sendtokenhelp'` key, as I do not possess the skills necessary to translate the new string into these languages.

I made a simple change to the `sendtoken.twig` template that checks the value of the `$mail_address_use_ldap` configuration variable. When it is enabled, it presents the new translated string, when it is disabled, it presents the previous translated string.

The English string that I added is as such:

> Enter your user name to reset your password. An email will be sent to the address associated with the supplied user name. When you receive this email, click the link inside to complete the password reset.

Feel free to provide any feedback on this PR. I am more than willing to make changes as desired by the development team.


Jitsusama:
Thanks for the PR @plewin. As you can see, it's been merged into my fork.

coudot:
Thanks for the patch @Jitsusama !

@plewin I let you merge the PR if this is ok for you.

PR #202, opened 2018-05-17
include config.inc.local.php + warning

nqb:
Same config done in white-pages (see https://github.com/ltb-project/white-pages/issues/30).

Maybe fixes #72 

- [ ] update doc

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -20,6 +20,11 @@
 #==============================================================================
 
 #==============================================================================
+# All the default values are kept here, you should not modify it but use
+# config.inc.local.php file instead to override the settings from here.
+#==============================================================================
+
+#==============================================================================
 # Configuration
 #==============================================================================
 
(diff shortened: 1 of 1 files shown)


coudot:
Thanks!

nqb:
Doc updated

conf/config.inc.php
```diff
@@ -288,3 +293,8 @@
 
 # Launch a posthook script after successful password change
 #$posthook = "/usr/share/self-service-password/posthook.sh";
+
+# Allow to override current settings with local configuration
+if (file_exists (dirname (__FILE__) . '/config.inc.local.php')) {
+    include dirname (__FILE__) . '/config.inc.local.php';
```
plewin: I would have preferred a _require_ instead of _include_, _require_ will stop the execution with an error (instead of a warning with continued execution) if the config.inc.local.php is not readable (or less likely, is a directory) which is potentially a future common issue if the file is not shipped by the package.

conf/config.inc.php
```diff
@@ -288,3 +293,8 @@
 
 # Launch a posthook script after successful password change
 #$posthook = "/usr/share/self-service-password/posthook.sh";
+
+# Allow to override current settings with local configuration
+if (file_exists (dirname (__FILE__) . '/config.inc.local.php')) {
```
plewin: dirname ( __ FILE __) is old syntax, __ DIR __ should be used (we are no longer supporting php 5.3 and older)

coudot:
@plewin I updated the code with your remarks

PR #208, opened 2018-06-13
Strict checking of input parameter values

Abdoulsore:
Need to check that the login, the newpassword, comfirmpassword and oldpassword given by the user have the right type (in this case: strings) before  going any further, to avoid issues with ldap_bind().

### DIFF
--- a/pages/change.php
+++ b/pages/change.php
@@ -35,6 +35,13 @@ $userdn = "";
 if (!isset($pwd_forbidden_chars)) { $pwd_forbidden_chars=""; }
 $mail = "";
 
+##verify that variables are strings
+
+if (isset($_POST["confirmpassword"]) and !is_string($_POST["confirmpassword"])) {return;}  
+if (isset($_POST["newpassword"]) and !is_string($_POST["newpassword"])) {return;} 
+if (isset($_POST["oldpassword"]) and !is_string($_POST["oldpassword"])) {return ;}
+if (isset($_REQUEST["login"]) and !is_string($_REQUEST["login"])) {return;}
+
 if (isset($_POST["confirmpassword"]) and $_POST["confirmpassword"]) { $confirmpassword = $_POST["confirmpassword"]; }
  else { $result = "confirmpasswordrequired"; }
 if (isset($_POST["newpassword"]) and $_POST["newpassword"]) { $newpassword = $_POST["newpassword"]; }


coudot:
Hello,

we indeed have an issue if $oldpassword is an array. There is a problem with php-ldap which is not setting errno after ldap_bind in this case. I will report it to PHP team.

For the workaround, using is_string may not be the better. I would prefer set a non zero value to $errno before calling ldap_bind, as the main issue is linked to ldap_bind method.

coudot:
Bug open on php.net, but marked as private for now as it can be a major security flaw: https://bugs.php.net/bug.php?id=76469

coudot:
PHP developers propose to use:
```
declare(strict_types=1);
```

davidcoutadeur:
We could use this in our code, but it should also be done in php-ldap, isn't it?

coudot:
I think we can not rely on ldap_errno. I will see if we can first check if ldap_bind was successful before checking errno.

coudot:
Seems we indeed need to update our code to accept bind if $errno is zero and $bind is success. Might be indeed a bug in our code.

coudot:
So the safer is to first test the return code of ldap_bind (true/false) before calling errno.

This is my proposed patch:
```
diff --git a/pages/change.php b/pages/change.php
index 1247788..abf504b 100644
--- a/pages/change.php
+++ b/pages/change.php
@@ -127,6 +127,7 @@ if ( $result === "" ) {
 
     # Bind with old password
     $bind = ldap_bind($ldap, $userdn, $oldpassword);
+    if (!$bind) { $result = "badcredentials"; }
     $errno = ldap_errno($ldap);
     if ( ($errno == 49) && $ad_mode ) {
         if ( ldap_get_option($ldap, 0x0032, $extended_error) ) {
```

Is it ok for you ?

davidcoutadeur:
Ok, seems efficient.
However, for security reasons, I still think it could be good to stop the process if we detect invalid format in these variables.
Whatever the result of php-ldap functions is, we must consider in ssp code that it is a failure.

coudot:
Agree, we can as a check on input to validate we have strings. But instead of a simple return which will display a blank page, we should display an error message.
And the check should be done on all pages.

Would you like to do it? I can aldo work on it.

Abdoulsore:
Yes, it is OK for me and I also agree with @davidcoutadeur on checking the validity of entries.

davidcoutadeur:
> But instead of a simple return which will display a blank page, we should display an error message.

I am not sure to agree with that.
- Entering a non matching old password is a mistake. It is ok, and we should display errors
- Entering an array as oldpassword is an attempt of attack, and we should not return anything back.

coudot:
Indeed, we can do that. @plewin what do you think?

plewin:
ldap_connect, ldap_set_option, ldap_get_values, ldap_first_entry, ldap_modify_batch etc all need to have return code checked too to be rigorous

Sadly proper handling of ldap functions returns is hard, I like to read zend ldap code to remember that.
See how zend framework do to handle ldap return codes to 
https://github.com/zendframework/zend-ldap/blob/master/src/Ldap.php#L115
https://github.com/zendframework/zend-ldap/blob/master/src/Ldap.php#L140

I believe `if (!$bind) { $result = "badcredentials"; }` is a correct approach to fix the issue.

I do not think that checking for strings is necessary. 
We should return a HTTP bad request idealistically. I do not think it is worth investing time to add lines reject in this SSP version.
I would recommend type casting instead, `$confirmpassword = strval($_POST["confirmpassword"]);` or (string)
If there is a conversion there will be a notice and the variables will have proper type.

coudot:
Thanks @plewin for your feedback. Using strval() seems a good idea and don't need to put a return instruction in the code. I will close this PR and open a new issue to force string conversion on input values.

PR #224, opened 2018-07-13
SMS OVH provider

dgeo:
Added OVH's php-sms-api via composer
Switched ReCaptcha to use composer (simplifies vendor/autoload.php !)
Added an optional parameter to send_sms_by_api() to get provider's config


dgeo:
I switched recaptcha to composer because composer use did break existing one (overriding autoload.php)
Nos sure if this is how you want to manage vendors/ dir ?

dgeo:
well… I switched PHPMailer, ReCaptcha to composer too, with few changes on includes.
Tell me if you prefer a less-invasive change… (I tested recaptcha, phpmailer and - of course - OVH sms)
"vendor"'s code could also live alone (as git submodules or requiring "cd lib; composer install" after cloning)

coudot:
Thanks a lot for this pull request but it is really huge, I need time to review it.

coudot:
This PR is really too big, we need to split it. Introducing composer is quite a big change and requires a separate PR. This could be done after 1.4 release.

dgeo:
now this is as light as it can be ;)

coudot:
Including OVH API in composer is quite heavy, for a feature most user will not use. I'm not sure if we should have it by default.

dgeo:
I think you are right… could this work if I create another "composer.json" in a subdir with sms script ?
I'd like to be able to deploy this without patching existing files

dgeo:
(and I'm not a real dev… just hacking there)

coudot:
Maybe this could be a separated project? I will think about it.

dgeo:
now this can be taken apart. I didn't include vendor/, only composer.json and I completed the doc

dgeo:
You prefer that I maintain this in a separate project ? so we could close this PR ?

dgeo:
changes are still needed in htdocs/sendsms.php anyway…

coudot:
Hello,

your current solution seems ok: let administrators run composer to install required dependencies.

I will include it like for 1.4 release.

PR #225, opened 2018-07-16
rate-limiting with json files

dgeo:
Add rate-limiting capacity, per IP AND per login

This should avoid simple script-kiddies

Seems to work for me™, no warranty ;)

Errors are hidden, I think we should add error strings…


dgeo:
I removed them and added a "normal" error message (to be translated for !(en|fr)

dgeo:
using this patch for years here… feel free to improve

coudot:
Hello @dgeo 

I applied your modifications and fixed some files, could you see if all is ok?

It will be part of 1.4 release

PR #227, opened 2018-07-25
Change some strings to be more descriptive.

daniellockard:
I have had users not be aware what the 'Question' or 'Email' links do, often assuming it's for sending an email to support, or something like that.  I think these strings are a bit more descriptive.


plewin:
I am unsure about these changes,

We are aware of this issue, that is why we have now popovers to explain the links. cf. https://github.com/ltb-project/self-service-password/pull/121

If we were to update these links we should update 'SMS' and 'SSH Key' too.

I am in favor of letting what it is now and wait for a redesigned user experience, something I hope to bring in v2.0

daniellockard:
That's fine - The popovers don't "work" for most of my users, as they don't even think to mouse over them. I have en.inc.php overridden in my local docker container.

JuJuMDX:
Hello,

You can set your own text in the menu by adding a new file in the conf directory.

For your example, you can create a file conf/en.inc.php which contains your translations : 
```
$ cat conf/en.inc.php
<?php
$messages['menuquestions'] = "Reset Password Via Security Question";
```

PR #233, opened 2018-08-04
Translated some lines on file pt-BR.inc.php to Brazilian Portuguese

AlexsanderAR:
Just finished translation to Brazilian Portuguese.


coudot:
Thanks!

PR #239, opened 2018-09-10
Fix in_array() error

mfinelli:
In 6bc80dafa1292d76745ff509692530ee2438fa76 the
$obscure_failure_messages option was added but it doesn't have any
default value so it defaults to null. in_array() expects the second
parameter to be an array, so we explicity cast it to an array in the
event it hasn't been set or hasn't been set as an array.

This fixes the following warning:
PHP Warning: in_array() expects parameter 2 to be array, null given
in /usr/share/self-service-password/pages/sendtoken.php on line 212

Note that this notice still exists so, $obscure_failure_messages should
probably also be initialized to something valid:
PHP Notice:  Undefined variable: obscure_failure_messages in
/usr/share/self-service-password/pages/sendtoken.php on line 212

### DIFF
--- a/pages/change.php
+++ b/pages/change.php
@@ -183,7 +183,7 @@ if ( $result === "" ) {
 #==============================================================================
 # HTML
 #==============================================================================
-if ( in_array($result, $obscure_failure_messages) ) { $result = "badcredentials"; }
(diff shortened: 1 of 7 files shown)


coudot:
Thanks for this fix!

PR #255, opened 2018-11-23
Be able to use different accounts to perform password changes

bmfp:
Password change : bind with user credentials or $ldap_binddn / $ldap_bindpw
Password reset : bind with privileged account

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -38,6 +38,8 @@ $ldap_url = "ldap://localhost";
 $ldap_starttls = false;
 $ldap_binddn = "cn=manager,dc=example,dc=com";
 $ldap_bindpw = "secret";
+$ldap_reset_binddn = "cn=privilegedaccount,dc=example,dc=com";
+$ldap_reset_bindpw = "terces";
 $ldap_base = "dc=example,dc=com";
 $ldap_login_attribute = "uid";
 $ldap_fullname_attribute = "cn";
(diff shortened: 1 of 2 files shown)


coudot:
Hello,

I don't see the need to have $ldap_binddn  and $ldap_resetbinddn. Could you explain the use case?

bmfp:
With current behavior, if $ldap_binddn and $ldap_bindpw are set, these credentials are used to change user's password (pages/change.php) or to reset user's password (pages/resetbytoken.php) ; if empty, user's credentials are used.
I wanted to be able to allow a user to make the change operation with his own credentials (empty $ldap_binddn and $ldap_bindpw), and use a configured privileged account to perform reset operation (with $ldap_reset_binddn / $ldap_reset_bindpw)

BR

coudot:
Is it not already possible with the `$ŵho_change_password` configuration parameter ?

See https://ltb-project.org/documentation/self-service-password/latest/config_ldap#credentials

cicob:
I guess I'd do +1 on this PR.
If however the abovementioned flag `$who_change_password` can actually be used, then we're all good.

Can someone please explain (and document) how this flag is supposed to work? 
I have tried to ask in #271 but I've so far nto seen any light in the tunnel...

plewin:
Hum, I checked the code, the assertion `if empty, user's credentials are used.` is incorrect.
Yes, setting ŵho_change_password = user would solve the presented use case.

A valid use case for this PR is wanting to use anonymous binding for user search and at the same time a service account for password reset. This is a valid use case but I am unsure it's op use case and if a pull request for this should be accepted.

coudot:
I agree, after reading this discussion, I think the use of `ŵho_change_password` is enough to solve this issue, the PR is not needed.

PR #264, opened 2019-02-03
add support for password files (aka pathway to docker secrets)

r2evans:
When using SSP in a docker environment, it may be convenient (and more secure) to not hard-code the SMTP and LDAP passwords in envvars. One solution is to support password *files*, where the password is stored in a file managed by the docker environment. With `docker-compose` (i.e., in a docker swarm), this still mandates an actual file on the docker host, but with k8s it can be done with cli-managed secrets that need-not be stored in a file.

This PR adds one function `get_passfile_or_pass` that takes two arguments: the original variable and an optional filename. If the original variable is defined, it is returned quickly with no further logic, so the default case of using this version changes nothing.

It also adds two variables: `smtp_pass_file` and `ldap_bindpw_file` to go along with the original variables. If `smtp_pass` is blank/unset and `smtp_pass_file` is set, then it reads the password directly from the file (verbatim, text-read, the file should not have a trailing newline) and uses that instead.

The rest of this PR tries to add the call to this function before any attempted use of the password.


coudot:
Hello @r2evans 

thanks for this proposition, but I don't see the need of adding this method in the code. Your idea is very good but we could just call `get_file_contents` from the config.inc.php, like:
```php
$ldap_bindpw = get_file_contents("/tmp/secret")
#$ldap_bindpw = "secret";
```

I can add this in the documentation.

coudot:
Great, we are working on our side on docker images for SSP, we should publish them soon.

r2evans:
(You meant `file_get_contents`, right?)

coudot:
Yes indeed.

http://php.net/manual/en/function.file-get-contents.php

PR #270, opened 2019-03-15
Added initial Norwegian bokmål (nb-NO) translation

thnilsen:
I have done an inital Norwegian (bokmål) translation that might be of interest to others.


coudot:
Thanks a lot!

PR #272, opened 2019-03-15
Allowing Email as URL-Parameter

cicob:
Thanks for offering this nice application!

You currently allow the user-ID to be fed via URL. With this amendment you can also add the email-address in the URL as in this example:

     https://SERVERNAME/?action=sendtoken&login=jw&usermail=john.wayne@hollywood.com   

When fetching the latest stable release from the LTB-project we always add these few rows before deploying the apllication to our portal-server. It is very convenient for the new users when we send out welcome-emails. They just click on the link to start the change-process. Both fields are already populated with ID and email.

C


pages/sendtoken.php
```diff
@@ -31,11 +31,26 @@
...
 $userdn = "";
 $token = "";
+$usermail = "";
 
 if (!$mail_address_use_ldap) {
-    if (isset($_POST["mail"]) and $_POST["mail"]) { $mail = strval($_POST["mail"]); }
-     else { $result = "mailrequired"; }
-}
+    if (isset($_POST["mail"]) and $_POST["mail"]) {
+        $mail = $_POST["mail"];
```
plewin: Can you use `strval(` for consistency please.

pages/sendtoken.php
```diff
@@ -31,11 +31,26 @@
...
+        
+    } elseif (isset($_REQUEST["usermail"]) and $_REQUEST["usermail"]) {
+        ###Check if email-address is submitted via URL
+        $usermail = $_REQUEST["usermail"];
```
plewin: Same here, can you use `strval(` for consistency please.

plewin:
@cicob thank you for the suggested changes, I think it's a nice addition, I suggested some modifications in the PR for consistency with the rest of the code.

cicob:
Super. Nice alignment to the rest of the code.
Thanks!

pages/sendtoken.php
coudot: Done in #411

pages/sendtoken.php
coudot: Done in #411

PR #274, opened 2019-03-18
Add base64 encoding option to passwords in posthook commands

thnilsen:
This change will allow the posthook script to handle passwords with characters
that would normally be escaped, rendering a different password to what was
actually entered by the user. Posthook script will have to handle decoding of
base64 strings.

- New true/false config option: $posthook_password_encodebase64
- New function posthook_command($posthook, $login, $newpassword, $oldpassword = null, $posthook_password_encodebase64 = false, $use_oldpass = false)

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -299,6 +299,9 @@ $default_action = "change";
 # Launch a posthook script after successful password change
 #$posthook = "/usr/share/self-service-password/posthook.sh";
 #$display_posthook_error = true;
+# Encode passwords sent to posthook script as base64. This will prevent alteration of the passwords if set to true.
+# To read the actual password in the posthook script, use a base64_decode function/tool
+#$posthook_password_encodebase64 = false;
 
 # Hide some messages to not disclose sensitive information
 # These messages will be replaced by badcredentials error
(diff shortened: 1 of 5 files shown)


coudot:
I don't understand why you need to add the `$use_oldpass` parameter, can't we just test if `$oldpassword` is null in the function?

thnilsen:
The reason for $use_oldpass parameter is to handle cases where the old password is blank, and the user is changing the password, not resetting it. If we just did a check for is_null/empty on old_pass, the routine would call reset rather than change.

coudot:
From my point of view, the differnce between change/reset should be managed into the posthook script itself.
Our documentation does not mention that the oldpassword is sent only in change mode. We send it anyway, even it is null: https://ltb-project.org/documentation/self-service-password/latest/config_posthook

thnilsen:
Your current master code does differ between a change and reset, as with reset it only sends the new password. Which is why I added to flag to differentiate. 

https://github.com/ltb-project/self-service-password/blob/039804ebc46ffe8e77f37f6f94e7e53af8e0fb76/pages/resetbytoken.php#L188

In any case I will go ahead and make the change and remove the flag for $oldpassword.

coudot:
Indeed, the old password is not sent in reset, but you can indeed have the same behavior in the function, by not sending the old_password if it is null. That's why I think the flag is not useful.

coudot:
Thanks for the changes!

PR #296, opened 2019-05-13
add policy: disallow special character at beginning or end

pisys:
It's a common strategy for users to fulfill the "special char" requirement by just appending a special char to the password.

This policy prevents this by checking if there is exactly 1 special character used and if it is at the beginning or end.


coudot:
All seems fine, thanks for this pull request!

coudot:
Doc updated: https://ltb-project.org/documentation/self-service-password/1.4/config_ppolicy#characters

PR #298, opened 2019-05-14
Disambiguate user by mail

pisys:
When sending token: If `ldap_search` returns multiple entries for a given username (ie. `$ldap_login_attribute` is not unique), then check all users and their mail addresses for verifying user input.


coudot:
I am not surewe should implement this. There should not be duplicate user identifiers in the directory.

pisys:
Since it is allowed to set any `$ldap_filter`, this feature would be a direct consequence from that, I think. After all, the `mail` attribute is unique anyways.

Eg. in my case I want to enable users to initiate password reset even if they don't know their exact username (The did not pick the username themselves, instead it was generated automatically from their common name). So my `$ldap_filter` looks like `(cn=*{login}*)`.

Together with the mail address provided (which is unique) I can verify that user input is correct.

However, I now spotted another problem with `resetbytoken.php`. There the user is fetched from LDAP by the `login` value, which is stored in the session and equals to what the user entered in the input form. Here it is not possible to apply the same logic.

So I'd improve this pull request in this way: 

* Store all user data, which was fetched from LDAP in `sendtoken.php`, in the session.
* Retrieve user data from session in `resetbytoken.php` instead of fetching it from LDAP again. 
* Display the value of `$ldap_login_attribute` in the password change form.

Issue #44, opened 2016-10-21
Illegal passwords list


aurbjo:
It would be great to be able to specify a list of illegal passwords in the config file, or even illegal words in password.

r2evans:
(Caveat: I'm not a dev with LTB.) Are you talking about a simple blacklist? It sounds like you mean substring matches, meaning `!password123!` should fail, since either `password` or (at least) `password123` should not be a substring.

On a cursory search, there are API-based subscription services (such as https://www.passwordrbl.com). I think this is different than what you are suggesting (and considerably different to implement).

I assume you mean a "static" blacklist ("static" in that its updates are either manual (by the sysadmin) or manual/versioned (by the maintainers of LTB-SSP). There are several known lists (such as http://web.archive.org/web/20130605082350/http://www.isdpodcast.com:80/resources/62k-common-passwords/, among many), so finding a reliable and reasonable source should just take a little time to research. I don't have much experience looking for them, so that link is just an example; I can't speak to it.

Some discussion questions:
- [ ] should this be case-insensitive?
- [ ] should it look for (simple) `1337`-speak transliteration?
- [ ] what about looking for reversals? (`321wrodssap`)
- [ ] if substring matching is done, is there a point at which leading/trailing characters make it alright?
(difference between `aaPASSWORD123aa` and `&13as!&8sd@@PASSWORD123@@f7y*sd`)

Obviously the combinations of all this needs to be considered for performance: a great password-matching blacklist does nothing if it takes too long to process and the page times out.

I suggest a first cut on this would be literal string matching, with later extensions to include substrings (with padding limits) and reversals.

What do you think, @coudot, do you find utility in having a blacklist? (I am confident that a lot of my authenticating-users will curse at me for enabling this :-)

@aurbjo, if he likes the idea (and doesn't have a starting point), it might help if you found an acceptable starting point for the list. Think: freely distributed, possibly updated (not antiquated), small enough, and easily parsed (perhaps just a literal text file, one-per-line).

coudot:
Yes, we could indeed use a simple text file as a blacklist, this file could then be updated by any means outside SSP. We can provide a simple blacklist at installation.

r2evans:
As a thought, would it be a bit robust to use something like [cracklib](http://php.net/manual/en/intro.crack.php) for checking more than just a dictionary? It supports more than just blacklists, though I admit I don't know how to configure its various checks.

aurbjo:
[...]

PR #299, opened 2019-05-15
New policy: Forbidden words

pisys:
Depending on the context or domain, there may be words which are more likely to be used by users in their passwords. This policy prevents that passwords contain any of the strings defined in `$pwd_forbidden_words`.


coudot:
See issue #44 

There is a discussion on how implement this feature

But your PR can be a first step. The best would be to use a file containing forbidden words

PR #302, opened 2019-05-15
Override policy rules if password shorter than a certain value

pisys:
One may define weak policy rules for long passwords (encouraging passphrases) and override these weak rules by harder ones if the user wants to choose a shorter password.

See the example in the config file.

This change should be completely backwards compatible.


coudot:
Hello @pisys,

thanks for all contributions on the software. I will need some time to read the code and give some feedback.

Can I ask how you will use Self Service Password? It seems you need a lot of new features.

pisys:
I use it for an association's LDAP directory. There are a couple of services which depend on it. Self Service Password is a great piece of software for managing just everything around passwords!

PR #303, opened 2019-05-16
add support for setting multiple question/answers.

shawnmckinney:
This updated setquestions.ph makes a few changes. It still does not require that the question attribute be multivalued, but if it is, now supports adding > 1.

1. The search for object class now pulls back the question attribute.
2. If the attribute is present in the search results...
  for each attribute do:
      decrypt question/value (if need be)
      compare the question itself to the question the user want to change.
      if there is a match, perform a ldap_mod_del on the existing attribute.
3. Add the new question/answer.

### DIFF
--- a/pages/setquestions.php
+++ b/pages/setquestions.php
@@ -153,10 +153,12 @@ if ( $result === "" ) {
     $answer_value = '{'.$question.'}'.$answer;
     $i = 0;
     $answers[$i++] = $crypt_answers ? encrypt($answer_value, $keyphrase) : $answer_value;
+    # Do we need to process multiple answers on this request?
     if ($aValues != NULL && $multiple_answers == true ) {
         #  Now determine if this answer has already been registered. If yes, don't add to the answer array.
         $question = preg_quote($question,'/');
         $pattern  = "/^\{$question\}(.*)$/i";
+        # Examine each previous question registered and look for matches.
         foreach ( $aValues as $key => $answer_encrypt ) {
             $value = $crypt_answers ? decrypt($answer_encrypt, $keyphrase) : $answer_encrypt;
             if (!preg_match($pattern, $value)) {
--- a/pages/setquestions.php
+++ b/pages/setquestions.php
@@ -88,7 +88,7 @@ if ( $result === "" ) {
     # Search for user
     $ldap_filter = str_replace("{login}", $login, $ldap_filter);
     $search = ldap_search($ldap, $ldap_base, $ldap_filter);
-
+    
     $errno = ldap_errno($ldap);
     if ( $errno ) {
         $result = "ldaperror";
@@ -124,8 +124,8 @@ if ( $result === "" ) {
         $bind = ldap_bind($ldap, $ldap_binddn, $ldap_bindpw);
     }
 
-    # Check objectClass presence
-    $search = ldap_search($ldap, $userdn, "(objectClass=*)", array("objectClass") );
+    # Check objectClass presence and pull back previous answers.
+    $search = ldap_search($ldap, $userdn, "(objectClass=*)", array("objectClass", $answer_attribute) );
  
     $errno = ldap_errno($ldap);
     if ( $errno ) {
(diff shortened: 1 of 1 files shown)


shawnmckinney:
There are a couple of problems with the code.

1. This isn't needed:
        //$answer   = preg_quote($answer,'/');
2. Need to verify this call was successful:
                $del = ldap_mod_del($ldap, $userdn , $deldata);
and maybe 
3. Do the ldap_mod_del and ldap_mod_add in a batch?
ldap_modify_batch

coudot:
You can also get all values in the code, replace the answer if this is needed, and then do a replace of all values. This will work in all cases.

shawnmckinney:
Well yeah, that's a much better approach.  OK, made the change, please have a look and see if this will work better.

shawnmckinney:
Ah crap, left off a check to see if multiple answers is supported.  Thinking we'll need another config parameter to specify:
$multiple_answers = true;

That way the code will know if replace or add when an answer already present.  Still thinking about it...

coudot:
Thanks, I will look at it as soon as possible

shawnmckinney:
Thanks. I've tested with multi and single-value attributes.  Seems to work OK, but had to add a config var to handle both cases:
$multiple_answers = true;

We discussed this earlier but prolly will need something like this too, for verification:
$number_answers = 2;

I think we can eliminate the first one via the second.  That is, if a single value attribute is enabled, no additional config variables must be enabled.  For multi-valued, this must be there, even if set to 1.
$number_answers = 1;

shawnmckinney:
As I think more about implementing the answer counter, I'm starting to have doubts that it can be done easily.  I suppose it wouldn't be too difficult to pin N sets of question/answer fields onto the page where n = $number_answers.

But this isn't very user friendly approach.  If user gets one wrong, they must renter them all one again on the next try.

The alternative is to have a page flow that prompts the user for each successive question/answer until they either a. have too many failures or b. achieve the correct number of answers.  This would introduce state into the application and so that's why I have my doubts.

There must be some other way to do it but this is all I can come up with so far.

--Shawn

shawnmckinney:
OK never mind about supporting the multiple answers.  It's not a requirements at this time so want to avoid the complexity.  So, having multiple questions, and one answer is good enough.

shawnmckinney:
Hi,

It's been some time since there's been any activity on this PR.  Is this still under consideration?  What are the next steps to get this code accepted?  Please advise.

Thanks,
--Shawn

pages/setquestions.php
```diff
@@ -88,7 +88,7 @@
     # Search for user
     $ldap_filter = str_replace("{login}", $login, $ldap_filter);
     $search = ldap_search($ldap, $ldap_base, $ldap_filter);
-
+    
     $errno = ldap_errno($ldap);
```
coudot: Please do not add spaces here

pages/setquestions.php
```diff
@@ -136,33 +136,45 @@
     # Get objectClass values from user entry
     $entry = ldap_first_entry($ldap, $search);
     $ocValues = ldap_get_values($ldap, $entry, "objectClass");
-
     # Remove 'count' key
```
coudot: Please do not remove blank line

pages/setquestions.php
```diff
@@ -136,33 +136,45 @@
...
     unset($ocValues["count"]);
 
+    $aValues = ldap_get_values($ldap, $entry, $answer_attribute );
+    # Remove 'count' key
+    unset($aValues["count"]);
+    
     if (! in_array( $answer_objectClass, $ocValues ) ) {
-
```
coudot: Please do not remove blank line

pages/setquestions.php
```diff
@@ -136,33 +136,45 @@
...
+            $value = $crypt_answers ? decrypt($answer_encrypt, $keyphrase) : $answer_encrypt;
+            if (!preg_match($pattern, $value)) {
+                $answers[$i++] = $answer_encrypt;
+            }
+        }
+    }
+    $userdata[$answer_attribute] = $answers;
     $replace = ldap_mod_replace($ldap, $userdn , $userdata);
-    
     $errno = ldap_errno($ldap);
```
coudot: Please do not remove blank line

pages/setquestions.php
```diff
@@ -136,33 +136,45 @@
...
     if ( $errno ) {
         $result = "answermoderror";
         error_log("LDAP - Modify answer (error $errno (".ldap_error($ldap).")");
     } else {
         $result = "answerchanged";
     }
-
```
coudot: Please do not remove blank line

coudot:
Hello @shawnmckinney 

sorry, I had no time to work on it. I made some cosmetics remarks than you can fix easily.

For the code, your proposal allows to set multiple answers, but did you then test if any of this answer can be used to reset the password?

You also need to declare the new configuration parameter in config.inc.php

shawnmckinney:
> On Jul 22, 2019, at 11:42 AM, Clément OUDOT <notifications@github.com> wrote:
> 
> sorry, I had no time to work on it. I made some cosmetics remarks than you can fix easily.

Hello Clement,

No problem, we’re all volunteers here.  :-)

One question, I’m not well versed in the protocol here.  

I saw your request for changes.  I accept all of the cosmetic changes.  Is this something I need to then do (in code), or can I perform an operation where your suggestions are automatically inserted?

> 
> On Jul 22, 2019, at 11:42 AM, Clément OUDOT <notifications@github.com> wrote:
> 
> For the code, your proposal allows to set multiple answers, but did you then test if any of this answer can be used to reset the password?

Yes, I did quite a bit of testing.  Stepping through with the debugger.  Everything appears to work but it wouldn’t hurt for another set of eyes, in case I missed something due to inexperience.

> 
> On Jul 22, 2019, at 11:42 AM, Clément OUDOT <notifications@github.com> wrote:
> 
> You also need to declare the new configuration parameter in config.inc.php

OK, will do.

Thanks,
—Shawn

coudot:
Hello @shawnmckinney 

thanks a lot for your work, I merged your code.

shawnmckinney:
You’re welcome.  Thanks for the help and guidance.  

—Shawn

Issue #97, opened 2017-03-24
Add an easy way to override messages


coudot:
The goal is to be able to easily override or add a new message.

coudot:
Done. Now we can just have a conf/$lang.inc.php that will have customized messages.

PR #309, opened 2019-06-06
the user should understand that the LDAP password change takes time

arno01:
I've seen people reading the message "Your password was changed" thought that it was while it wasn't since many LDAP clients cache the previous password and it takes time for the cache to be refreshed.


coudot:
Sorry, I can't accept this change, as this is not true for evey installation. LDAP client should not cache credentials.

In next version, you will be able to customize the message by adding a lang file in conf/, see #97

arno01:
Yeah, that makes sense to use the custom lang files, thanks for the pointer!

> LDAP client should not cache credentials.

Why is that? In the non-redundant setups when the LDAP server is offline for the maintenance or just died, the credentials caching is the only thing that saves the users.
And some LDAP setups actually do prefer to cache the passwords in order to reduce the load in case if there are many clients or the network is flaky.

Example https://linux.die.net/man/5/sssd.conf

```
cache_credentials (bool)
  Determines if user credentials are also cached in the local LDB cache
  User credentials are stored in a SHA512 hash, not in plaintext
```

PR #321, opened 2019-07-30
Several tweaks we always do manually

rolffokkens-th:


coudot:
To be studied after 1.4 release

stevleibelt:
A question for the `pwd_static_policy` change. Wouldn't it be "cleaner" if we use smarty tpl to not implement another way of rendering templates?
Where will come the time, where the content of "pwd_static_policy" will contain some if/then-else parts.

Most important question, is this sill open/relevant or should we close it?

Cheers,
Stev

Issue #109, opened 2017-05-08
Use Smarty framework


Kage:
I've become a big fan of the [Silex PHP framework](https://silex.sensiolabs.org), and was wondering what your thoughts are on rewriting it to fit?  That way it has a proper router, dependency injection, etc., and dependencies can be easily managed with Composer.

I have a FOSS project I'll be publishing soon on [my community](https://github.com/HackThisCode) that could also work as an excellent skeleton framework to merge this into.

So.. thoughts?

coudot:
Why not. For LTB White Pages I used Smarty which is also a good solution.

plewin:
I believe SSP should implement the first PSRs (coding standard, code style, logging, autoloader) and refactor most of the code before thinking about porting to a microframework or even any framework.
This could be targetted for 1.2.

coudot:
Yes, there is a lot of work before this. I think if we achieve to rewrite SSP with a framework, we could release it as 2.0. For the moment we will try to release 1.1 and add new features in 1.x branch.

ntoniazzi:
Silex has [officially been deprecated](http://symfony.com/blog/the-end-of-silex), so another framework would be needed (Symfony).

plewin:
Yes silex is deprecated, I work on an new major version based on symphony
components

Kage:
Indeed.  I would recommend [Slim Framework](https://www.slimframework.com) instead of a full-fledged Symfony app for two reasons:

1. Symfony 3.4 is quite cumbersome, and Symfony 4+ requires PHP 7.2 (thus severely limiting the audience for SSP).  Integrating SSP into either would probably be quite onerous and require a complete recode of everything.
2. Slim allows for basic framework control (router, PSR by default), but is very lightweight and would probably allow for reusing a lot of existing SSP code.

I've noticed Smarty mentioned elsewhere, and I would instead strongly recommend using Twig for templating.  Smarty is also cumbersome, whereas Twig is quite lightweight and has much more community support.

coudot:
I used Smarty for LTB White Pages and it would be better for me to have the same framework for SSP

Kage:
Fair enough. :P

coudot:
Work done by @JuJuMDX, a PR is coming

PR #332, opened 2019-08-31
Docker file

halkeye:
First pass at a docker file. I think everything is working right. I'll be pushing and deploying to my local install later today, but all quick tests work.

Used the official php images, but added the modules I could tell/read are needed.

Didn't update travis or anything else cause I assume hub.docker.com is easier than maintaining your own support.

I mostly want tagged releases so its easy to upgrade 

And all the various docker images floating around that I could find replace  conf/config.inc.php so conf/config.inc.local.php doesn't work and you have to define every variable.


halkeye:
I made  a small change to fix compiling curl module, but i'm running it locally and everything seems to be working fine. So its ready for review!

coudot:
Thanks a lot @halkeye we will have a look on it.

coudot:
Hello @halkeye 

Thanks a lot for this PR. I tested the build, it works well.

How do you configure the conf.inc.local.php in this image?

halkeye:
Externally mapping it to a local dir is the easiest

```
docker run -it --name ssp --rm -v /path/on/local/config.inc.local.php:/var/www/html/conf/config.inc.local.php ssp-docker-image
```

or they can use this as a base image and add it to their own.

```
FROM ltb-project/self-service-password:0.3
COPY ./config.inc.local.php /var/www/html/conf/config.inc.local.php
```

I always recommend dependabot in this case so it'll send PRs to update the version number in your docker file

incaseoftrouble:
Hi,

Not sure where to put this, this MR seems to be a reasonable place. I'm using a separate php-fpm container (:alpine), the following Dockerfile (based on the one in this MR) works for me. I tested sending mail & changing password. Since I'm enabling the same php extensions as the other Dockerfile, I suppose they should provide the same functionality.

`FROM php:fpm-alpine
RUN apk --update --no-cache add curl libldap icu-libs libbz2 && \
    apk --no-cache add --virtual .build-deps curl-dev openldap-dev icu-dev bzip2-dev && \
    docker-php-ext-install -j$(nproc) bcmath bz2 iconv intl mbstring opcache curl ldap && \
    apk del .build-deps
`
I guess it would be easiest to include this somewhere at the bottom of some README.md / Docker.md :)

halkeye:
So i've pushed this up to https://hub.docker.com/repository/docker/halkeye/self-service-password if anyone wants to use is.

I still think its worth merging into the repo so it can have tagged releases.

halkeye:
@coudot  is there anything else you want me to do about this?

coudot:
Hello @halkeye 

seems ok, I wait also for other reviews before merging this. This is targeted for next version, but no due date for the moment.

encryptblockr:
this needs documentation in README.md
@halkeye i think you are missing documentation to show people how to use this with docker
you can add that in README.md file

encryptblockr:
@halkeye i coned your branch and build docker image from it without changing anything and here is error message am getting

```
 Token encryption requires a random string in keyphrase setting
```

image

coudot:
> @halkeye i coned your branch and build docker image from it without changing anything and here is error message am getting
> 
> ```
>  Token encryption requires a random string in keyphrase setting
> ```

It just mean that you need to change the default keyphrase in configuration.

ztj1993:
This feature is very good, it is recommended to merge as soon as possible, and the configuration file supports the way to use environment variables.

halkeye:
Super basic helm chart for those using kubernetes - https://hub.helm.sh/charts/halkeye/ldap-self-service (uses my fork of the image, since there's no official one)

domsgit:
Why this useful pr has passed three months and has not been merged??? highly recommended

domsgit:
> Why this useful pr has passed three months and has not been merged??? highly recommended
@soisik @maxbes

coudot:
Hello @halkeye 

we included Smarty framework in 1.4 version (see #109)

Would it be possible to update your Docker image to have Smarty in it?

halkeye:
um, I've ended up switching to keycloak, so I don't end up using this anymore.

Whats required to get this done? 

Did a bunch of digging into the code, looks like /usr/share/php/smarty3/ needs to have smarty according to https://github.com/ltb-project/self-service-password/blob/97b341ceefce8bda831c19de77e5fac850c4d623/conf/config.inc.php#L337-L339

so something like: 
```
RUN wget -O - https://github.com/smarty-php/smarty/archive/v3.1.35.tar.gz | tar xzf - -C /usr/share/php/smarty3/ --strip-components=1
```

?

if you tell me what needs to be done i can update the PR

coudot:
Maybe just:
```
apt-get install smarty3 smarty-gettext
```

halkeye:
🤞

coudot:
Thanks but my suggestion was maybe not accurate, as using Debian packages is not working:
```
The following packages have unmet dependencies:
 smarty3 : Depends: php or
                    php-cgi or
                    php-cli
           Depends: php-common but it is not installable
```
As php is not installed with package, we should do the same for Smarty

halkeye:
So my original suggestion of extracting the zip to the folder? I don't know how to install smarty. The few times I've used PHP recently they have composer setup for dependancies

coudot:
Your solution seems fine indeed

As I am not a Docker expert, I do not really understand why php is not installed as package.

halkeye:
It's compiled. https://github.com/docker-library/php/blob/master/7.2/stretch/apache/Dockerfile#L166

Probably so they can provide different versions than whatever Debian provides

halkeye:
Okay quick test it runs.

I think you should merge it, and then do any fixes later as a new PR

coudot:
Hello @halkeye 

I tested this morning, and all seems fine, thanks a lot for your help!

Issue #331, opened 2019-08-31
SMS Twilio Integration


halkeye:
I think I'm going to right a twilio client for ssp, Its been a long time since I wrote straight up php, but it looks like https://www.twilio.com/docs/libraries/php is pretty easy to use.

Is this something that would be interested upstream in the project? or I should keep it local?

coudot:
Hello @halkeye 

For now, we have a wiki page to provide code samples for differents API providers: https://ltb-project.org/documentation/self-service-password/latest/sms_api

PR #333, opened 2019-09-01
SMS Twilio Integration

halkeye:
Looks like you vendored the other libraries, so i just grabbed https://github.com/twilio/twilio-php/archive/master.zip and committed it.

Seems to work for me. Is there a standard way to report errors back? should I be using error_log maybe?

### DIFF
--- a/lib/smsapi-twilo.inc.php
+++ b/lib/smsapi-twilo.inc.php
@@ -1,0 +1,101 @@
+<?php
+#==============================================================================
+# LTB Self Service Password
+#
+# Copyright (C) 2019 LTB-project.org
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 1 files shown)


coudot:
Hello @halkeye 

Thanks for the PR. But importing 1600+ files in the project is not a good idea, even if it is a vendor library. I don't want to extend the code base, we should find a way either to let users download the library, either to choose only the relevant files needed.

halkeye:
@coudot thats a super super good point, and waking up this morning the answer seems obvious.

This is just a single api call, so i'm going to revert ea63cbb and just use curl (like the one mentioned in #331).

on #331 you mention the wiki page, would that be preferred? or is a simple script here fine.

coudot:
I think we can do both: provide the sample script and a documentation on the wiki

halkeye:
Okay, new version, handful of lines.

Added a bunch of notes as the comment on the top on what to setup/how to use

```php
/*
 * Add to your config.inc.local.php
 * Credentials are on the main page of the dashboard - https://www.twilio.com/console
 * Phone numbers can be purchased and managed from the phone number dashboard
 * - https://www.twilio.com/console/phone-numbers/search
 *
 * $twilio_sid = '<sid>';
 * $twilio_auth_token = '<authtoken>';
 * $twilio_outgoing_number = '+18881234567';
 * true if you want to ask twilio if the number is valid, and reformat it
 * false otherwise
 * $twilio_lookup_first = true / false
 */
```

coudot:
Doc: https://ltb-project.org/documentation/self-service-password/1.4/sms_api#twilio

PR #333 merged

Thanks!

halkeye:
Awesome.

You may want to add the other config values to the wiki as well

coudot:
Done

PR #336, opened 2019-09-05
Translate Pwned, specialatends and logic

donjoh:
Translated Pwned, specialatends and made some text more logical


coudot:
Thanks!

PR #342, opened 2019-10-02
Hide token URLs unless debug mode is on

fjannot:
* Tested for mail page, not on SMS one. 
 * Having full token URL into log could be a security lack. Disable full URL logging unless debug mode is enabled.


coudot:
Thanks!

PR #344, opened 2019-10-05
Composer + Bootstrap 4 + New Icon

mrdrogdrog:
I made this fork a long time ago. It changes some of the core behaviour like:
- Usage of composer for libs
- Replace Bootstrap 3 with Bootstrap 4
- New Logo/Icon


coudot:
Thanks for your proposal. Some remarks:
* why did you remove badges from README?
* unit tests are failing because of crypto API

mrdrogdrog:
I removed the badges, because when I wrote this fork, I didn't think about creating a PR. I can add them again. That's no problem.

I can fix the unit tests too, but first I wanted to know if you like the new design. Also this PR contains a breaking change, because it can't be used with PHP < 7.2. pwned-passwords 2.0 requires this version.

coudot:
I did not see that, it can indeed be a blocker as this software can be used on systems with older PHP versions.

Anyway, using bootsrap 4 is something I wanted to do, but before I think I will first try to solve #150 to have HTML templates.

It would then be hard to accept your PR directly, as it indeed involves a lot of changes. It can be interesting for a major version, but we are currently working on 1.4 release, which should not break anything regarding current 1.3 version.

mrdrogdrog:
I didn't expect this PR to be merged in one piece because it's affecting multiple parts of the code. It's more of an idea, what it can look like.
Feel free to use pieces of my code base to improve the UI / add composer!

I know that backwards compatibility is important, but PHP 5 is deprecated as hell and I think it's a good idea to drop it.

coudot:
THanks for this, I will indeed keep this PR and look at your code when we will decide to use composer and bootstrap4.

coudot:
To be reworked and merged after 1.4 release

coudot:
Seems all has been done on our side in 1.6 release

PR #347, opened 2019-10-18
Option to deploy with docker

cmachadox:
I added the option to be able to deploy the application with the docker, created a dockerfile and add the description in README


coudot:
Hello, how is it diifferent from #332 ?

PR #355, opened 2019-11-14
ltb-project#354 : Can use change tab when default_action not set to c…

JuJuMDX:


coudot:
Indeed, it works also with:
```php
$use_change = false;
```

PR #360, opened 2019-12-12
Improved pt-BR.inc.php with more colloquial form

bkmeneguello:
Also sorted entries to match "en.inc.php" ordering to ease translations


coudot:
Thanks a lot!

PR #388, opened 2020-05-20
changing accents to htmlentities code

Thelvaen:
some of the accents were in htmlentities format, some were directly typed in. As it mess things up when I had to edit it within a shell box, I changed everything in the french language to the proper htmlentities codes.


coudot:
Hello @Thelvaen 

Source files are in UTF-8, you should be able to edit them from your shell with the proper language settings.

We should remove HTML entities and use UTF8 characters in lang files

Thelvaen:
Ok, sounds good to me, one way or another is good to me :)

Also I'm trying to add other complexity checks, but I might not be clear on how to use git for those operation ^^

PR #389, opened 2020-05-20
ADD: Samba synchronization via call to smbpasswd

davide84:
This commit solves the problem of synchronizing the Samba passwords with those
 of LDAP, when the two tools are used in parallel. The 'standard' way of
 storing Samba information in LDAP seems not to be working anymore in Samba4.

The assumptions are as follows:
 - we are on a machine where smbpasswd command is available;
 - users are created in both LDAP and Samba with the same passwords

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -63,10 +63,6 @@ $samba_mode = false;
 #$samba_options['max_age'] = 45;
 #$samba_options['expire_days'] = 90;
 
-# use smbpasswd to synchronize samba password with LDAP
-# (!) Warning: it assumes both LDAP and Samba have the same accounts and passwords
-$use_smbpasswd = false
-
 # Shadow options - require shadowAccount objectClass
 # Update shadowLastChange
 $shadow_options['update_shadowLastChange'] = false;
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -63,6 +63,10 @@ $samba_mode = false;
 #$samba_options['max_age'] = 45;
 #$samba_options['expire_days'] = 90;
 
+# use smbpasswd to synchronize samba password with LDAP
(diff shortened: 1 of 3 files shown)


coudot:
Would this be not better to use postscript hook for this?

Another point, you just call the command when password is changed with oldpassword, but not if it is reset.

davide84:
Hi! You're right, scripts are a better place (I didn't notice that feature at first).
I updated the branch, the only diff now is the additional file in scripts/. You might still want to add it to the project as a working example (I tested it on my server). It took me one day of searches and experiments before finding that working solution.
By the way, self-service-password is probably the easiest and nicest piece of my LDAP/Samba installation... thanks :)

coudot:
Hello @davide84 

thanks for this contribution. I think to configuration option is no more needed. I can merge the sample script.

For my curiosity, did you try to enable Active Directory mode? It should work with Samba4.

davide84:
I tried to configure Samba as AD DC, hoping to drop OpenLDAP and centralize everything. But first I got issues with the services, then port conflicts (for the records, Ubuntu 20.04). I would have preferred that, but in the end it was just way simpler to have two independent services and manually sync the passwords...
It's not that elegant, I admit... just s confirmation in case I try again, with that configuration would the password be uniquely stored, and changeable via SSP?

coudot:
> I tried to configure Samba as AD DC, hoping to drop OpenLDAP and centralize everything. But first I got issues with the services, then port conflicts (for the records, Ubuntu 20.04). I would have preferred that, but in the end it was just way simpler to have two independent services and manually sync the passwords...
> It's not that elegant, I admit... just s confirmation in case I try again, with that configuration would the password be uniquely stored, and changeable via SSP?

I am not a Samba expert, but I think that using only the Samba LDAP directory should work, and configuring SSP with AD mode against it is enough.

PR #395, opened 2020-06-10
Feature enhancements to security question functionality

mtkraai:
This pull request includes four distinct changes in three commits:

1. Add config setting to disable the built-in questions. This makes it easier to completely replace the built-in security questions without editing any language files.  Use cases include containerization and configuration management systems, where a single configuration file is helpful.

2. Add config setting to store multiple answers in a single string value.  AD doesn't support child objects of User objects the way the default settings work.  This change allows multiple answers to be stored in a single string, such as the comment attribute as suggested by the documentation, or one of the "custom/extension" attributes.

3. Add config setting to require multiple answers to change a password.  We have a requirement that three security questions be answered before resetting the password.  By setting the count > 1, the required number of questions are displayed on both the "set questions" page, and the "reset by questions" page.  All of the questions must be set (or answered correctly) for the operation to proceed.

4. Add config setting to allow users to retrieve their registered questions on the "reset by questions" page.  If the user wants to be reminded which question(s) he answered, submitting the form with only the login name will retrieve the questions, leaving only the answers blank.

I have made an effort to ensure that all changes are backward compatible.  If the new settings are left at their defaults, the existing behavior will not change.


coudot:
There are already accepted PR for multiple answers: #303 and #318 

Could you see what are the differences? And maybe only propose enhancements to previous PR.

mtkraai:
My changes come after those PRs were merged, and are enhancements or related additional features.

#318 looks like it's just the missing configuration variable for #303.

The changes in #303 don't work in Active Directory.  Commit 86e7363 adds an alternate storage mechanism for the multiple answers, which does work in AD.

#303 allows multiple answers to be set, one at a time.  With commit 66f9c06, a certain number of multiple answers can be _required_, and are all on one page.  The same setting requires the same number of questions to be answered correctly before the password is reset.

I can create separate PRs for each of these new features, if you'd like.

coudot:
No it's ok, thanks for the details. We will have a look at it.

coudot:
See #456

Issue #404, opened 2020-07-28
consider to move inline resources


juniper275:
Hi,

i'm testing some CSP policy and i noticed that some inline resources must be moved to their own file:

Line 15

'<style>
--
  | html, body {
  | background: url("images/unsplash-tim.jpeg") no-repeat center fixed;
  | background-size: cover;
  | }
  | </style>'

Line 77

'<script>
--
  | $(document).ready(function(){
  | // Menu links popovers
  | $('[data-toggle="menu-popover"]').popover({
  | trigger: 'hover',
  | placement: 'bottom',
  | container: 'body' // Allows the popover to be larger than the menu button
  | });
  | });
  | </script>'

plewin:
Hi, thanks for the suggestion,
I opened a PR for the script part. https://github.com/ltb-project/self-service-password/pull/406
For the style part, I don't see how to this easily because it conflicts with the customizable background feature.
I don't think the lack of support of CSP style-src, it would be nice to have.

plewin:
This issue is resolved for 1.4 with this PR https://github.com/ltb-project/self-service-password/pull/406
SSP will work with Content Security Policy (CSP) for javascript

The customizable background feature conflicts with the CSP rule _style-src_. If one desires to implement it, they should disable the background-image or move the css to their own css file.

maglub:
The CSP issue crept back in already 2021, in the templates/footer.tpl. When I use CSP, I get an error. The issue is in both templates/header.tpl, and templates/footer.tpl.

```
16:45 $ git blame templates/footer.tpl
311da283 (Julien VEDRINE  2020-02-17 13:47:27 +0100  1) </div>
311da283 (Julien VEDRINE  2020-02-17 13:47:27 +0100  2) {if $display_footer}
311da283 (Julien VEDRINE  2020-02-17 13:47:27 +0100  3) <div id="footer">LDAP Tool Box Self Service Password - version {$version}</div>
311da283 (Julien VEDRINE  2020-02-17 13:47:27 +0100  4) {/if}
45254838 (David Coutadeur 2024-03-12 10:26:35 +0100  5) <script src="vendor/jquery/jquery.min.js"></script>
311da283 (Julien VEDRINE  2020-02-17 13:47:27 +0100  6) <script src="vendor/bootstrap/js/bootstrap.min.js"></script>
0f79d3b7 (Clément OUDOT   2020-08-10 18:54:00 +0200  7) <script src="js/self-service-password.js"></script>
[...]

PR #406, opened 2020-08-04
Move inline javascript to its own file fixes (#404)

plewin:
Move javascript to its own file to allow the use of HTTP Content-Security-Policy script-src

ref https://github.com/ltb-project/self-service-password/issues/404


coudot:
Thanks for this, you can merge it !

Issue #220, opened 2018-07-10
Pre Hook script


coudot:
The goal is to call a script before calling the password change. This would allow to check the password against any API or script.

PR #416, opened 2020-08-27
Prehook - ltb-project/self-service-password#220

faust64:


coudot:
resolve #220

coudot:
I added an option to ignore prehook error, which will be false by default.

Thanks a lot for the contribution @faust64 !

PR #419, opened 2020-09-08
Update config_ldap.rst

PGTBoos:
Previously  in the bind string cn=manager was used, this let to confusion here, as we used the cn value also in $who_change_password.
Now i made it more clear that this attribute is more of a boolean kind it can have two values only.

BTW... please update the website (https://ltb-project.org/documentation/self-service-password) as well to point to this manual,  only after raising this issue i was pointed to this wiki. currently this gip repository forwards to the website for the manual

### DIFF
--- a/docs/config_ldap.rst
+++ b/docs/config_ldap.rst
@@ -59,19 +59,20 @@ LDAP server, you can try to set this option:
 Credentials
 -----------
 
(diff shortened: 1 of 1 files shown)


coudot:
Thanks for the contribution!

PR #428, opened 2020-10-18
Added best practices of autocomplete for password managers

ZenHarbinger:
https://web.dev/sign-in-form-best-practices/#new-password

Updated the templates/change.tpl file so that the current and new password entries use auto-complete to help password managers.


coudot:
Thanks for this PR. We could also apply it to other forms where we put the new password.

coudot:
Merged with few more commits

Issue #306, opened 2019-05-20
feature request: add another password quality check


codesmaker:
Hi,

First of all, thanks a lot for this very helpful piece of software.

I would like to request a feature where passwords quality is checked to make sure that users don't use the same password with only one character changed.

For example, if a user's password is "Myprecious#", a user can just change one character and it would be accepted just as "Myprecious@".

This check is implemented in pwquality and it is called "difok". This is the explanation provided in its config file:
```
##Number of characters in the new password that must not be present in the old password.
difok = 5
```

coudot:
Note that this rule can only be applied when user change its password by giving the old password. It won't work for password reset (by token, questions, etc.)

codesmaker:
Awesome. I understand.

Many thanks.

coudot:
Done by @faust64

PR #430, opened 2020-10-20
feat(diff-check): #306 

faust64:
Implements a password quality check. that was suggested by @codesmaker in ltb-project/self-service-password#306

When `$pwd_diff_last_min_chars` is `> 0`, old password and new password are compared, making sure there is at least `$pwd_diff_last_min_chars` new unique characters in the new password, that were not in the previous one.

Default value in config.inc.php is set to 0, which disables that check.


coudot:
Thanks a lot for this contribution!

PR #433, opened 2020-10-25
fix(branding): logo in menubar

faust64:
looking at whitepages (https://github.com/ltb-project/white-pages/blob/master/templates/menu.tpl#L13) and servicedesk (https://github.com/ltb-project/service-desk/blob/master/templates/menu.tpl#L13), there should be a logo in the menu bar.

also: removes a duplicate smarty assign (already done a few lines above the one I'm removing: https://github.com/ltb-project/self-service-password/blob/579e16c9f82cd80aeb19d229ec697b03ba587648/htdocs/index.php#L167)


coudot:
Thanks!

PR #456, opened 2021-01-12
Mtkraai master

JuJuMDX:
https://github.com/ltb-project/self-service-password/pull/395


coudot:
Thanks a lot to @mtkraai and @JuJuMDX 

Code seems ok, but we need to update our bulit-in documentation before merging

mtkraai:
I saw this activity this morning, and just yesterday I noticed that something is missing in my original PR: jquery.selectunique.js is loaded and called before the main jquery file, which causes it to fail.

I solved it by moving `<script src="js/jquery-3.3.1.min.js"></script>` to the bottom of the HTML head: https://github.com/mtkraai/self-service-password/commit/28af83d4ce18c49e7df2002a4f74cfc05034cb08

I know there are reasons for putting the JS at the bottom, and there are other ways to fix the issue, so I leave it up to you.

coudot:
I think the best would be to keep all js loading in the footer. We do this load quicker the HTML content. I'll this this point before merging.

coudot:
Other issue: there some code logic in lang files, I don't want this, we need a better way to manage it. And in current code, only english lang was modified.

coudot:
Other issue, `$questions_count` is not assigned in smarty template, so we never see multiple answers in form.

coudot:
> Other issue, `$questions_count` is not assigned in smarty template, so we never see multiple answers in form.

Fixed in https://github.com/ltb-project/self-service-password/pull/456/commits/c342f77a2cd8ed0852f4af451d51c9d535c31637

coudot:
> I saw this activity this morning, and just yesterday I noticed that something is missing in my original PR: jquery.selectunique.js is loaded and called before the main jquery file, which causes it to fail.
> 
> I solved it by moving `<script src="js/jquery-3.3.1.min.js"></script>` to the bottom of the HTML head: [mtkraai@28af83d](https://github.com/mtkraai/self-service-password/commit/28af83d4ce18c49e7df2002a4f74cfc05034cb08)
> 
> I know there are reasons for putting the JS at the bottom, and there are other ways to fix the issue, so I leave it up to you.

Fixed in https://github.com/ltb-project/self-service-password/pull/456/commits/c342f77a2cd8ed0852f4af451d51c9d535c31637

htdocs/setquestions.php
```diff
@@ -152,23 +179,33 @@
         $userdata["objectClass"] = $ocValues;
     }
 
-    $answer_value = '{'.$question.'}'.$answer;
     $i = 0;
-    $answers[$i++] = $crypt_answers ? encrypt($answer_value, $keyphrase) : $answer_value;
+    error_log(print_r($question, true));
+    error_log(print_r($answer, true));
```
plewin, line 184:
```suggestion
```

Are these forgotten debug lines ?

htdocs/resetbyquestions.php
```diff
@@ -132,23 +164,49 @@
...
         }
-    }
 
-    if (!$match) {
-        $result = "answernomatch";
-        error_log("Answer does not match question for user $login");
+        $matched = 0;
+        # Match answer(s)
+        for ($q = 0; $q < $questions_count; $q++) {
+            if ($registered_questions[$question[$q]] === $answer[$q]) {
```
plewin, line 201: hash_equals is recommended when comparing secrets
```suggestion
            if (hash_equals($registered_questions[$question[$q]], $answer[$q])) {
```

Issue #460, opened 2021-01-17
update from PHP-7.2


r2evans:
The `Dockerfile` pulls from php:7.2-apache, and PHP-7.2 is beyond its end-of-life (https://www.php.net/supported-versions.php).

The `.travis.yml` file suggests this package has not been tested against PHP newer than 7.2.

r2evans:
If it matters, I get the same results testing with PHP-7.2 and PHP-7.4.

Using docker php:7.2-apache and php:7.4-apache, and https://phar.phpunit.de/phpunit-5.7.phar, I ran `phpunit tests` in both. Granted, it doesn't pass, but the pass seems more likely to be related to a missing PHP package in the base `php` docker image ... and fails the same way in both.

This is not an authoritative test, but is encouraging.

(BTW: I've been running it on PHP:7.3 for a couple of years.)

## PHP-7.2

```bash
$ docker run -v /.../self-service-password/:/app/ -w /app/ --rm php:7.2-apache   /app/phpunit tests
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

....crypto: decryption error Integrity check failed.
.F.﻿﻿...                                                        10 / 10 (100%)

Time: 2.68 seconds, Memory: 10.00MB

There was 1 failure:

1) HookTest::testCommandWithSpecialCharacters
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'/usr/local/bin/hook.sh 'jdoé' 'jæne' 'jøhn''
+'/usr/local/bin/hook.sh 'jdo' 'jne' 'jhn''

/app/tests/HookTest.php:31

FAILURES!
Tests: 10, Assertions: 36, Failures: 1.
```

## PHP-7.3

```bash
$ docker run -v /.../self-service-password/:/app/ -w /app/ --rm php:7.4-apache   /app/phpunit tests
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

....crypto: decryption error Integrity check failed.
.F.﻿﻿...                                                        10 / 10 (100%)

Time: 2.06 seconds, Memory: 10.00MB

There was 1 failure:

1) HookTest::testCommandWithSpecialCharacters
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'/usr/local/bin/hook.sh 'jdoé' 'jæne' 'jøhn''
+'/usr/local/bin/hook.sh 'jdo' 'jne' 'jhn''

[...]

PR #468, opened 2021-02-22
updating php to 7.4

faust64:
Following up on https://github.com/ltb-project/self-service-password/issues/460


coudot:
Thanks @faust64 !

Issue #474, opened 2021-03-02
Language selection issue


spalekg:
If I configure:
```
$lang = "rs";
$allowed_lang = array("rs");
```
It should ignore language from the browser and it will be shown Serbian (rs) language. But this always shows the English language instead.

I took a look into lib/detectbrowserlanguage.php and there is no logic about using **$allowed_lang** and also in htdocs/index.php, there is no using **$allowed_lang** at all. Then I have a search in all PHP files about using **$allowed_lang** and still, there is not any place where it is used.

This should be fixed, so when I want to be shown ONLY one language (no matter of browser language or whatever) it should be possible.

coudot:
Indeed, this should have be broken by our switch to Smarty. We need to fix it.

PR #476, opened 2021-03-03
fix(lang): re-include allowed_lang check

faust64:
copy/pasta from previous branch.
see https://github.com/ltb-project/self-service-password/issues/474


coudot:
Tested on my side, it works

spalekg (commented):
In line 41 (**htdocs/index.php**):

``` 
if ($entry_lang == $lang || empty($allowed_lang) || in_array($entry_lang, $allowed_lang) ) {
```

Where is **$entry_lang** defined? It is not used, so why you put it in this statement?
Maybe that was defined somewhere before so you just copy/paste? Please fix it.

coudot:
Thanks for the review

Issue #447, opened 2020-12-09
Provide WebServices / REST API


coudot:
I would like to have some WebServices in Self Service Password.

## Configuration

These web services would not be enabled by default, so we need to create configuration parameters to enable them (a parameter for each service to enable)

### checkpassword

URL: `/rest/v1/checkpassword.php`

Input parameters:
* login (optional)
* oldpassword (optional)
* newpassword (required)

Call ` check_password_strength` internal method.

Result as JSON:
* error: 0 if new password is OK, 1 else
* result: result key (ex: `pwned`)
* message: default language message (ex: `Your new password has already been published on leaks, you should consider changing it on any other service that it is in use`)

### changepassword

URL: `/rest/v1/changepassword.php`

Input parameters:
* login (required)
* oldpassword (required)
* newpassword (required)

Call pre_hook, `change_password`, post_hook and mail notification.

Result as JSON:
* error: 0 if password is changed, 1 else
* result: result key (ex: `passworderror`)
* message: default language message (ex: `Password was refused by the LDAP directory`)

### adminchangepassword

URL: `/rest/v1/adminchangepassword.php`

Input parameters:
* login (required)
* newpassword (required)

Call pre_hook, `change_password`, post_hook and mail notification.

Result as JSON:
* error: 0 if password is changed, 1 else
* result: result key (ex: `passworderror`)
* message: default language message (ex: `Password was refused by the LDAP directory`)

coudot:
See 486aa44c7576773f9c079c49d9b1d82a557f4f03

coudot:
Still needed:
* Documentation
* Apache configuration

PR #483, opened 2021-03-08
add hook in rest api and a script for multi ldap change password

JuJuMDX:
#447

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -44,8 +44,6 @@ $ldap_fullname_attribute = "cn";
 $ldap_filter = "(&(objectClass=person)($ldap_login_attribute={login}))";
 $ldap_use_exop_passwd = false;
 $ldap_use_ppolicy_control = false;
-//$secondaries_ldap[0]['ldap_url'] = 'ldap://ldap2.example.com';
-//$secondaries_ldap[1]['ldap_url'] = 'ldap://ldap3.example.com';
 
 # Active Directory mode
 # true: use unicodePwd as password field
@@ -62,7 +60,6 @@ $ad_options['change_expired_password'] = false;
 # true: update sambaNTpassword and sambaPwdLastSet attributes too
 # false: just update the password
 $samba_mode = false;
-$samba_options = null;
 # Set password min/max age in Samba attributes
 #$samba_options['min_age'] = 5;
 #$samba_options['max_age'] = 45;
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -44,6 +44,8 @@ $ldap_fullname_attribute = "cn";
 $ldap_filter = "(&(objectClass=person)($ldap_login_attribute={login}))";
 $ldap_use_exop_passwd = false;
 $ldap_use_ppolicy_control = false;
+//$secondaries_ldap[0]['ldap_url'] = 'ldap://ldap2.example.com';
+//$secondaries_ldap[1]['ldap_url'] = 'ldap://ldap3.example.com';
 
 # Active Directory mode
 # true: use unicodePwd as password field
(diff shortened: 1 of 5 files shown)


conf/config.inc.php
```diff
@@ -44,6 +44,8 @@
 $ldap_filter = "(&(objectClass=person)($ldap_login_attribute={login}))";
 $ldap_use_exop_passwd = false;
 $ldap_use_ppolicy_control = false;
+//$secondaries_ldap[0]['ldap_url'] = 'ldap://ldap2.example.com';
+//$secondaries_ldap[1]['ldap_url'] = 'ldap://ldap3.example.com';
```
coudot, line 48: I don't think we should add this in config.inc.php

We can maybe add documentation about this script in doc/

conf/config.inc.php
```diff
@@ -60,6 +62,7 @@
 # true: update sambaNTpassword and sambaPwdLastSet attributes too
 # false: just update the password
 $samba_mode = false;
+$samba_options = null;
```
coudot, line 65: What is the need for this ?

scripts/multi_ldap_change.php
```diff
@@ -0,0 +1,135 @@
...
+
+$result = "";
+$return = Array();
+$error_code = 1;
+$login = $argv[1];
+$newpassword = $argv[2];
+$oldpassword = '';
+
+$log_file = fopen(__DIR__.'/multi_ldap_change.log', 'a+');
+fwrite($log_file, "Change '$login' password...\n");
```
coudot, line 20: We should not log in scripts/ directory by default.

Use instead https://www.php.net/manual/fr/function.sys-get-temp-dir.php

scripts/multi_ldap_change.php
```diff
@@ -0,0 +1,135 @@
...
+        fwrite($log_file, "LDAP - Unable to use StartTLS");
+    } else {
+
+    # Bind
+    if ( isset($ldap_binddn) && isset($ldap_bindpw) ) {
+        $bind = ldap_bind($ldap, $ldap_binddn, $ldap_bindpw);
+    } else {
+        $bind = ldap_bind($ldap);
+    }
+
```
coudot, line 38: You suppose here that connection settings to secondary repositories are the same that the main LDAP server.

Maybe your config object  `$s_ldap` should contain all connection parameters.

scripts/multi_ldap_change.php
```diff
@@ -0,0 +1,135 @@
...
+        fwrite($log_file, "LDAP - User $login not found");
+    } else {
+
+    # Get user email for notification
+    if ( $notify_on_change ) {
+        $mailValues = ldap_get_values($ldap, $entry, $mail_attribute);
+        if ( $mailValues["count"] > 0 ) {
+            $mail = $mailValues[0];
+        }
+    }
```
coudot, line 72: I don't see the need here, as the notification should have been done when the password has been changed on the main LDAP directory

scripts/multi_ldap_change.php
```diff
@@ -0,0 +1,135 @@
...
+    # Check objectClass to allow samba and shadow updates
+    $ocValues = ldap_get_values($ldap, $entry, 'objectClass');
+    if ( !in_array( 'sambaSamAccount', $ocValues ) and !in_array( 'sambaSAMAccount', $ocValues ) ) {
+        $samba_mode = false;
+    }
+    if ( !in_array( 'shadowAccount', $ocValues ) ) {
+        $shadow_options['update_shadowLastChange'] = false;
+        $shadow_options['update_shadowExpire'] = false;
+    }
+
```
coudot, line 83: Maybe too complex to add this in this script. Because we don't know if secondary LDAP are Samba/AD/...

I think this script should just connect to the secondary LDAP, search user and change password.

coudot (changes requested):

coudot (approved):

coudot:
Thanks a lot @JuJuMDX !

PR #501, opened 2021-04-22
Remove extra semicolon from setquestions template

spike77453:
This fixes an erroneous extra semicolon in the setquestions template (`index.php?action=setquestions`):

Screenshot_20210423_000850


coudot:
Hi @spike77453 

thanks a lot for the fix!

PR #502, opened 2021-04-23
Remove alt text so empty logo doesn't show 'msg_title' twice

spike77453:
Here's another tiny fix: This removes the `alt` attribute from the logo's `img` tag. If the logo is empty `msg_title` will show twice otherwise, e.g.

Screenshot_20210423_114704

### DIFF
--- a/templates/menu.tpl
+++ b/templates/menu.tpl
@@ -10,7 +10,7 @@
               </button>
               <a class="navbar-brand" href="index.php{if $default_action != 'change'}?action=change{/if}">
                 {if $logo}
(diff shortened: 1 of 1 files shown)


coudot:
It's not a good practice to remove `alt`attributes. We should instead remove the `img` markup if logo is not defined.

coudot:
Thanks!

spike77453:
> It's not a good practice to remove `alt`attributes. We should instead remove the `img` markup if logo is not defined.

You're absolutely right, I've changed it accordingly. I also altered the `alt` attribute to it matches `templates/index.tpl` (which I think is less confusing if the image actually doesn't load and `alt` is shown). Hope that's ok.

Issue #504, opened 2021-04-28
Cannot use docker get gregwar/captcha----use docker 


fuyangqing:
I use docker to Installat self-service-password,but find it's not have php-dg . Uncaught Error: Call to undefined function Gregwar\\Captcha

logs is 

2021-04-28T14:02:39.911938782Z [Wed Apr 28 14:02:39.911669 2021] [php7:error] [pid 18] [client *:49283] PHP Fatal error:  Uncaught Error: Call to undefined function Gregwar\\Captcha\\imagecreatetruecolor() in /var/www/lib/vendor/gregwar/captcha/src/Gregwar/Captcha/CaptchaBuilder.php:423\nStack trace:\n#0 /var/www/html/captcha.php(30): Gregwar\\Captcha\\CaptchaBuilder->build()\n#1 {main}\n  thrown in /var/www/lib/vendor/gregwar/captcha/src/Gregwar/Captcha/CaptchaBuilder.php on line 423, referer: http://192.168.*:443/?action=sendsms
2021-04-28T14:02:39.911971876Z * - - [28/Apr/2021:14:02:39 +0000] "GET /captcha.php HTTP/1.1" 500 307 "http://192.168.*:443/?action=sendsms" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.82 Safari/537.36 Edg/89.0.774.50"

Please help me 0.0

faust64:
It seems like the gd library is missing.
Testing the image built from this repository, I can reproduce the issue.
Could you confirm: which image are you using? The `docker.io/ltbproject/self-service-password:latest` one?

I'll submit a PR ...

fuyangqing:
> 似乎gd库丢失了。
> 测试从该存储库构建的映像，我可以重现该问题。
> 您能否确认：您使用的是哪个图像？在`docker.io/ltbproject/self-service-password:latest`一个？
> 
> 我将提交公关...



> It seems like the gd library is missing.
> Testing the image built from this repository, I can reproduce the issue.
> Could you confirm: which image are you using? The `docker.io/ltbproject/self-service-password:latest` one?
> 
> I'll submit a PR ...

YES ，I use docker.io/ltbproject/self-service-password:latest，and make docker run.
 emmm...   thanks

faust64:
Can you try to re-pull that image and restart your container?
[...]

PR #505, opened 2021-04-28
fix(captcha): missing gd library

faust64:
as reported in
https://github.com/ltb-project/self-service-password/issues/504


coudot:
Hi @faust64 

Dockerfile is now in packaging/docker

coudot:
Thanks @faust64

Issue #506, opened 2021-04-28
I have a little problem - I can't use SMS for the next step


fuyangqing:
Sorry, I'm here again...
Today,I deploy self-service，have a little problem
1. At present, I plan to use the service only through old passwords and SMS

2. There is no problem in using through old passwords ; But when I use the way of SMS: SMS can be received normally, but when I modify the password, it is similar to the old password. I try a lot of passwords, even if the password is completely different, it will be reported wrong. I also find different friends to help me try, but there is still a problem. So I want to ask you boss, what should I do

3.  I checked the log and found that there was no obvious error. I think there is something wrong with SMS file, but I can receive SMS normally，I downloaded the configuration file and SMS file, only hiding some key information
4. The following is the configuration file and related information
[smsapi.txt](https://github.com/ltb-project/self-service-password/files/6392718/smsapi.txt)
[111.txt](https://github.com/ltb-project/self-service-password/files/6392702/111.txt)
![1111](https://user-images.githubusercontent.com/34743908/116423647-74d8c200-a873-11eb-8b62-c365c99d7b20.PNG)
[config.txt](https://github.com/ltb-project/self-service-password/files/6392716/config.txt)

Regards,
A learner

coudot:
Maybe a bug.

Could you try this patch?
```
diff --git a/lib/functions.inc.php b/lib/functions.inc.php
index be55e6b..9506d0b 100644
--- a/lib/functions.inc.php
+++ b/lib/functions.inc.php
@@ -278,7 +278,7 @@ function check_password_strength( $password, $oldpassword, $pwd_policy_config, $
     # Same as login?
     if ( $pwd_diff_login and $password === $login ) { $result="sameaslogin"; }
 
-    if ( $pwd_diff_last_min_chars > 0 ) {
+    if ( $pwd_diff_last_min_chars > 0 and strlen($oldpassword) > 0 ) {
        $similarities = similar_text($oldpassword, $password);
        $check_len    = strlen($oldpassword) < strlen($password) ? strlen($oldpassword) : strlen($password);
        $new_chars    = $check_len - $similarities;
```

fuyangqing:
> Maybe a bug.
> 
> Could you try this patch?
> 
> ```
> diff --git a/lib/functions.inc.php b/lib/functions.inc.php
> index be55e6b..9506d0b 100644
> --- a/lib/functions.inc.php
> +++ b/lib/functions.inc.php
[...]

PR #507, opened 2021-04-28
fix(reset)

faust64:
as seen in #506


coudot:
Did you test the patch?

Maybe we should add a unit test.

faust64:
I did not
if `pwd_diff_last_min_chars` > 0 and `oldpassword` is an empty string, `similarities` is 0, `check_len` is 0, `new_chars` is 0. Reset would always fail.

coudot:
Thanks for the test

PR #513, opened 2021-05-05
fix(sshkeys): don't send mail notification when entry was not changed

faust64:


htdocs/changesshkey.php
```diff
@@ -148,7 +148,7 @@
 #==============================================================================
 # Notify password change
```
coudot, line 149: Could you also update comment?

faust64: sure!

faust64: ... and I managed to submit another one ... https://github.com/ltb-project/self-service-password/pull/542
not fixing comments ...

Issue #327, opened 2019-08-29
Configure several LDAP servers and select one depending on context


soisik:
Hi,

We are setting up a federalized identity management system where users will change their password in their local LDAP server using a federalized SSP. Currently, the only solution would be to install as many SSP instances as we have local LDAP server to change password in, and dispatch users to the right SSP instance when he request a password change. 

For many reasons, we would rather have one SSP instance to deal with, so we are proposing a new feature in SSP configuration management:

- Possibility to dynamically override global and local settings with an extra configuration file, whose reference key is passed in an HTTP_HEADER. Config files would be:
   - config.inc.php
   - config.inc.local.php (if exists)
   - config.inc.%extra-config%.php (if exists)

- HTTP Header name is `SSP-Extra-Config` by default, but can be overridden in config.inc.local.php.

For example, if an incoming request has the HTTP header `SSP-Extra-Config` with the value "mypartner", SSP will load the extra configuration file config.inc.mypartner.php.

PR following.

coudot:
Doc updated: https://ltb-project.org/documentation/self-service-password/1.4/config_general#multi_tenancy

jdavis7718:
I am unable to get this to work as documented and the documentation is a little lacking on it as well.
Here is where I am at with this.

If I name the file config.inc.local.php it is read in and display properly so I know my configuration in the file is correct.
I am attempting to use this with Active Directory so I named a config file config.inc.ad.php
I uncommented the line $header_name_extra_config = "SSP-Extra-Config:";  in config.inc.php to have it listen for that header.
In my apache config I added this line to always send the header equal to value ad
Header always set SSP-Extra-Config "ad"

Per your documentation when it sees the value ad it should read in the config.inc.ad.php file, it does not. 

On a Linux system I used wget to verify the header is added properly, output below
  Server: Apache/2.4.37 (Red Hat Enterprise Linux) OpenSSL/1.1.1g
  X-Powered-By: PHP/7.3.20
  SSP-Extra-Config: ad
  Keep-Alive: timeout=5, max=100
  Connection: Keep-Alive
  Transfer-Encoding: chunked
  Content-Type: text/html; charset=UTF-8

Any help would be much appreciated

faust64:
Hi,

Trying it out: the `Header always set ...` would not work. 
Instead, you should set: `RequestHeader setIfEmpty SSP-Extra-Config "ad"`

Let us know how that goes.

jdavis7718:
[...]

PR #515, opened 2021-05-05
docs(multi-tenancy): adds samples setting multi-tenancy header

faust64:
Following up on ltb-project/self-service-password#327

### DIFF
--- a/docs/config_general.rst
+++ b/docs/config_general.rst
@@ -38,6 +38,27 @@ This feature is disable by default. To enable it:
 Then if you send the header ``SSP-Extra-Config: mydomain``, the file
 ``conf/config.inc.mydomain.php`` will be loaded.
 
+Using Apache, we may set such header using the following:
+
+.. code:: apache
+
+    <VirtualHost *:80>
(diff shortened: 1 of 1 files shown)


faust64:
A sample based on the `SERVER_NAME` may make more sense - sticking with the existing webserver configuration.
Although in that case, we may want to rework the conf/config.inc.php, looking for both `HTTP_` prefixed and non-prefixed values.

docs/config_general.rst
```diff
@@ -38,6 +38,27 @@ This feature is disable by default. To enable it:
 Then if you send the header ``SSP-Extra-Config: mydomain``, the file
 ``conf/config.inc.mydomain.php`` will be loaded.
 
+Using Apache, we may set such header using the following:
+
+.. code:: apache
+
+    <VirtualHost *:80>
+       ServerName ssp.domain1.com
+       RequestHeader RequestHeader setIfEmpty SSP-Extra-Config domain1
```
coudot, line 47: RequestHeader is written twice

faust64: right...

PR #522, opened 2021-05-12
[Security:low] Dismiss captcha once it is used

coudot:
When captcha is used, it is deleted to it can not be used for other requests.


coudot:
@fuyangqing This fixes the issue on captcha that you reported

PR #536, opened 2021-05-27
adding Kerberos authentication support

jazzl0ver:
Point $ldap_krbccname to Kerberos auth cache file and comment out $ldap_bind* lines, for example:
```
$ldap_krbccname = "/tmp/krb5cc_apache";
```
Make sure the web server is able to read /tmp/krb5cc_apache.

/tmp/krb5cc_apache - is a Kerberos auth cache generated by k5start service from an AD user's keytab, like:
```
/usr/bin/k5start -f /usr/local/etc/aduser.keytab -b -a -K 120 -L -l 1d -k /tmp/krb5cc_apache -U -o apache -p /var/run/k5start_apache.pid
```


htdocs/change.php
```diff
@@ -89,6 +89,9 @@
         # Bind
         if ( isset($ldap_binddn) && isset($ldap_bindpw) ) {
             $bind = ldap_bind($ldap, $ldap_binddn, $ldap_bindpw);
+        } elseif ( isset($_SERVER['KRB5CCNAME']) ) {
+            putenv("KRB5CCNAME=".$_SERVER['KRB5CCNAME']);
+            $bind = ldap_sasl_bind($ldap, NULL, NULL, 'GSSAPI') or die('Failed to GSSAPI bind.');
```
coudot, line 94: You should not `die` here, the error will be catched further in the code.

jazzl0ver:
> Could you clarify the use case? The goal is that a user authenticated by Kerberos can change its password without entering its login and its old password?

The purpose of the patch is to allow a user to change its password w/o expressing admin credentials ($ldap_bindpw) in a text file and prevent somebody from steeling them.

coudot:
Ok, it is clearer now.

This should be added in documentation: https://github.com/ltb-project/self-service-password/blob/master/docs/config_ldap.rst#credentials

Could you add documentation changes in this PR?

coudot:
I don't understand why Kerberos authentication is needed in Apache configuration. Why does it not just work by setting `KRB5CCNAME` env var?

jazzl0ver:
You're right, I can add a config variable for krb5 cache file path instead

coudot:
Yes, it would be easier

coudot:
Thanks for your contribution!

adrian-agp:
One question: for this you have to join the ssp server to the domain, or you have to create a krb5.keytab file

davidcoutadeur:
> One question: for this you have to join the ssp server to the domain, or you have to create a krb5.keytab file

If I understand correctly, the service `k5start`, when starting, authenticates to a kerberos server (potentially an AD) thanks to the keytab (aduser.keytab), and get a TGT ticket.

Then ssp uses this ticket to authenticate to the AD.

adrian-agp:
k5start: error reading /usr/local/etc/aduser.keytab: Key table file '/usr/local/etc/aduser.keytab' not found

where do I find it???

adrian-agp:
es correcto esto que estoy haciendo 
ssp es el nombre de ususario y self es el nombre del servidor

samba-tool spn add ssp/self.example.com ssp
samba-tool domain exportkeytab /opt/SELF.keytab --principal ssp/self.example.com
rsync -avhp /opt/SELF.keytab root@192.168.20.5:/etc

/usr/bin/k5start -f /etc/SELF.keytab -b -a -K 120 -L -l 1d -k /tmp/krb5cc_www-data -U -o www-data -p /var/run/k5start_www-data.pid
k5start: error getting credentials: Client 'ssp/self.example.com@EXAMPLE.COM' not found in Kerberos database

PR #539, opened 2021-06-01
fix(version): mismatch between htdocs/index.php and rest/v1/include.php

faust64:
Either that, or set version in htdocs/index.php back to 1.4?


coudot:
I don't think we should use version in rest/v1/include.php, it should be removed.

Issue #538, opened 2021-05-31
Switch to a new Pwned Passwords module


coudot:
The module from ron-maxweb (https://packagist.org/packages/ron-maxweb/pwned-passwords) is abandoned.

We should use instead https://github.com/mxrxdxn/pwned-passwords

PR #540, opened 2021-06-01
Refactor pwned passwords

faust64:
As suggested in https://github.com/ltb-project/self-service-password/issues/538

Replaces ron-maxweb/pwned-passwords with mxrxdxn/pwned-passwords.

Tested locally with regular password modification, did not check API, resets, ... though I assume they would behave in a similar way.


lib/functions.inc.php
```diff
@@ -317,9 +317,9 @@ function check_password_strength( $password, $oldpassword, $pwd_policy_config, $
     }
 
     # pwned?
-    if ($use_pwnedpasswords) {
+    if ($use_pwnedpasswords and version_compare(PHP_VERSION, '7.2.5') >= 0) {
```
coudot, line 320: We should maybe add a test in htdocs/index.php to return "phpupgraderequired" if `$use_pwnedpasswords and version_compare(PHP_VERSION, '7.2.5') < 0`

rest/v1/include.php
```diff
@@ -16,7 +16,7 @@
 require_once("../../lib/vendor/defuse-crypto.phar");
 require_once("../../lib/functions.inc.php");
 if ($use_pwnedpasswords) {
-    require_once("../../lib/vendor/ron-maxweb/pwned-passwords/src/PwnedPasswords/PwnedPasswords.php");
+    require_once("../../lib/vendor/mxrxdxn/pwned-passwords/src/PwnedPasswords/PwnedPasswords.php");
```
coudot, line 19: Maybe in another PR, we should use composer also for webservices

lib/functions.inc.php
faust64: Will do

rest/v1/include.php
faust64: oh, and actually, the above probably won't work. As seen running Travis-CI tests, swapping one for the other would return an error regarding GruzzleHttp\Client not being loaded, which I fixed by calling the autoload.php instead, ... I'll fix this ...

PR #542, opened 2021-06-02
fix(notify): don't send notification if modification failed

faust64:
changing an ssh key, mail notifications would be issued, even when we failed to update keys in LDAP


coudot:
Nice catch

Issue #494, opened 2021-04-08
allow more than one mail_attribute value


muzzol:
on one of my setups I use two attributes with mail, one is optional (personalMail) and the other one mandatory (mail).

I would like to use first if found and then second one. something similar to ldap_url attribute:

    $mail_attribute = "personalMail mail";

thanks for your time

coudot:
Hello,

this is not possible now. It needs some code in htdocs/sendtoken.php 

See https://github.com/ltb-project/self-service-password/blob/master/htdocs/sendtoken.php#L120

muzzol:
I'm trying to configure this on a fresh test installation and I'm facing some problems.

https://self-service-password.readthedocs.io/en/latest/config_mail.html

docs here refers to **$mail_attributes** but in config.inc.php I only see **$mail_attribute** so I guess first one is a typo.

I can use **$mail_attribute = "mail"**; but if I try to configure an array like this  **$mail_attribute = array('personalMail', 'mail');** I get a mailnomatch error: **The email address does not match the submitted user name**

and this appears in logs:
```

[Thu Sep 09 12:28:08.304621 2021] [php7:warn] [pid 15739] [client 172.16.236.12:52060] PHP Warning:  ldap_get_values() expects parameter 3 to be string, array given in /var/www/html/ssp/htdocs/sendtoken.php on line 122, referer: http://xx.xx.xx.xx/ssp/index.php
[Thu Sep 09 12:28:08.304700 2021] [php7:warn] [pid 15739] [client 172.16.236.12:52060] PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /var/www/html/ssp/htdocs/sendtoken.php on line 138, referer: http://xx.xx.xx.xx/ssp/index.php
[Thu Sep 09 12:28:08.304731 2021] [php7:notice] [pid 15739] [client 172.16.236.12:52060] Mail not found for user XXXXX, referer: http://xx.xx.xx.xx/ssp/index.php
```

Is there any other configuration I'm missing?

coudot:
Are you using code from master?

muzzol:
no, just downloaded from GH:

`wget https://github.com/ltb-project/self-service-password/archive/refs/tags/v1.4.3.zip`

coudot:
As you can see, the fix is merged in master and will be available in 1.5

coudot:
@faust64 there will be a need for an upgrade note, as we changed a configuration attribute name

[...]

PR #546, opened 2021-06-06
feat(mails): using several mail attributes

faust64:
As discussed in ltb-project/self-service-password#494


muzzol:
thanks for your time!!

Issue #556, opened 2021-06-17
Somthing missing in the OVH API doc


zebux:
Hello In the documentation to the SMS API OVH : 
https://self-service-password.readthedocs.io/en/latest/sms_api.html#sms-api

I think somethink is missing. It is write : "Go to lib/ovhsms and type" but nothing after and then the configuration to set in the config file. So I think that something is missing.
Regards
Eric
![SSP-doc miss](https://user-images.githubusercontent.com/35032402/122396712-905d7080-cf78-11eb-9b33-902c79bb603b.png)

coudot:
Documentation fixed. Thanks a lot for the report.

PR #557, opened 2021-06-17
fix(docs): sms-api

faust64:
As reported in ltb-project/self-service-password#556


coudot:
I already fixed it. Thanks for the help

PR #558, opened 2021-06-17
fix(docs): ratelimit check interval should be 1h, not 1min

faust64:
Doc mentions ratelimit would last for 1h.
Then quotes the defaults from `config.inc.php`: `60s`.

Set it to 1h everywhere.


coudot:
I think the issue is in documentation, blocking authentication for 1 hour seems really too long. Documentation mentions a "short time" so it should be 1 minute and not 1 hour

coudot:
Thanks!

Issue #573, opened 2021-09-07
How to reset password via api


ilanni2460:
How to reset password via api

I don’t know if it’s written like this：

curl -H "Content-Type: application/x-www-form-urlencoded" -X POST -d '{ "login":"abc","newpassword": "Wer123456", "oldpassword":"Wquy507"}' http://ssp.ilanni.com/rest/v1/changepassword.php

faust64:
Note that you are setting a content type ton form-urlencoded while posting a JSON object.
First I would try to change the Content-Type, or reformat the posted data.

Let us know how that goes

ilanni2460:
@faust64 
I now use this to make a request, and no information is returned

curl -X POST -d '{ "login":"abc","newpassword": "Wer123456", "oldpassword":"Wquy507"}' http://ssp.ilanni.com/rest/v1/changepassword.php


The ssp error log is：

192.168.0.24 - - [08/Sep/2021:15:20:11 +0800] "POST /rest/v1/changepassword.php HTTP/1.1" 500 - "-" "curl/7.29.0"

ilanni2460:
@faust64 Actually I don’t know how to reset the user password through the curl command

can you help me,tks

faust64:
In JSON, should be something like:

```
curl -X POST https://xxx/rest/v1/changepassword.php
   -H 'Content-Type: application/json'
   -d '{"login":"username","oldpassword":"my_old_password","newpassword":"my_new_password"}'
```

Or keeping the form/url-encoded

```
curl -X POST https://xxx/rest/v1/changepassword.php
   -H 'Content-Type: application/x-www-form-urlencoded'
   -d 'login=username&oldpassword=my_old_password&newpassword=my_new_password'
```

If your error 500 persist with proper input:

- Make sure your configuration sets `$use_restapi = true;`.
- Make sure you are already able to change passwords, using web UI.

coudot:
Link to OpenAPI spec: https://github.com/ltb-project/self-service-password/blob/v1.4/rest/v1/doc/openapi-spec.yaml

ilanni2460:
@coudot @faust64 

i use the command:
`curl -X POST -H 'Content-Type: application/x-www-form-urlencoded' -d 'login=abc&oldpassword=Wquy507&newpassword=Wer123456' http://ssp.ilanni.com/rest/v1/changepassword.php
`

the ssp access log:
`
[...]

PR #576, opened 2021-09-09
fix(api): phpmailer needs to be included (#573)

faust64:
as reported by @ilanni2460 
When use_pwndpasswords is false, phpmailer init would result in error 500


coudot:
Thanks a lot Samuel

PR #583, opened 2021-09-13
add aliyum sms

ilanni2460:
add aliyum sms


coudot:
Hello @ilanni2460 

You should not push all composer dependencies in code.

See example of OVH API: https://self-service-password.readthedocs.io/en/latest/sms_api.html#ovh

We have a composer file in the lib/ folder, and run composer in it when needed.

Could you update your PR to conform to this?

Issue #602, opened 2021-11-07
Errors when Using Captchas with Password Reset Emails


tropf:
Dear maintainer,

thank you for creating and publishing this self service portal. It saved me from setting up a much more heavyweight solution.
My local setup produced a bug in the latest version (1.4.3) that I'd like to report:

## Description
When enabling captchas **and** password reset emails (`$use_token = true`),
the reset email is properly sent **but** errors occur:

```
Warning: ini_set(): A session is active. You cannot change the session module's ini settings at this time in /var/www/self-service-password/htdocs/sendtoken.php on line 174

Warning: ini_set(): Headers already sent. You cannot change the session module's ini settings at this time in /var/www/self-service-password/htdocs/sendtoken.php on line 175

Warning: session_name(): Cannot change session name when session is active in /var/www/self-service-password/htdocs/sendtoken.php on line 177
```

## Steps to reproduce
1. Setup "LDAP Tool Box Self Service Password" and connect it to a working ldap database
2. Make sure errors are printed to the web page (!)
3. Request a password reset link through the web interface for an existing user with a correct email address

## Expected Behavior
The email is sent and no errors are displayed.

## Actual Behavior
The email is sent and the link in it works, but the errors copy-pasted above are printed and visible on the web page.

## Notes
(I haven't written PHP for more than five years, so take this with a grain of salt.)

- Checking the captcha invokes `session_start()` (in the same file).
- This checking happens **before** the token is generated, which calls `session_name("token")` which throws the error because a already started session can't be renamed.
- Note that this call to `session_name()` fails, but the reset functionality is still available. (Does this introduce a security risk?)

faust64:
Hi,

See #603 for a possible fix / to test.
The session_name warning shouldn't be an issue: we would use the session ID, submitted in the confirmation e-mail, to confirm a request is legitimate.
The ini_set use_cookies overrides a default. AFAIU, ensures no cookie would be kept on client side. (https://www.php.net/manual/en/session.configuration.php#ini.session.use-cookies)
The ini_set use_only_cookies should ensure a PHP default wasn't changed: shouldn't be an issue, unless your php.ini says otherwise. AFAIU, ensures client won't be able to assume arbitrary sessions IDs. (https://www.php.net/manual/en/session.configuration.php#ini.session.use-only-cookies)

tropf:
Thank you for the patch.

The errors disappeared, but now the captcha verification does not work anymore.
[...]

PR #603, opened 2021-11-07
fix(sendtoken): session started twice when captcha enabled

faust64:
Trying to address ltb-project/self-service-password#602

to test

### DIFF
--- a/htdocs/sendtoken.php
+++ b/htdocs/sendtoken.php
@@ -47,12 +47,11 @@ if (!$mail_address_use_ldap) {
 }
 if ($use_captcha) {
     if (isset($_POST["captchaphrase"]) and $_POST["captchaphrase"]) { $captchaphrase = strval($_POST["captchaphrase"]); }
-     else { $result = "captcharequired"; }
+    else { $result = "captcharequired"; }
 }
 if (isset($_REQUEST["login"]) and $_REQUEST["login"]) { $login = strval($_REQUEST["login"]); }
- else { $result = "loginrequired"; }
-if (! isset($_POST["mail"]) and ! isset($_REQUEST["login"]))
- { $result = "emptysendtokenform"; }
+else { $result = "loginrequired"; }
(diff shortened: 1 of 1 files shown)


liedekef:
This fix is not ok according to me. Best practices:
- avoid using ini_set (it can be blocked, and you shouldn't even need to touch it for sessions, obey what the system tells you there)
- use session_start as early as possible (heck, maybe even already just after config.inc.php, when checking if token or sms are needed)
- avoid setting the session name if you have different scripts, just adds to possible errors
- if the need for calling session_start multiple times is there, check if not already started:
```
if (session_id() == "")
  session_start();
```
- for token: don't use sessions or cookies or the database at all for the reset link and the token, use nonces (I can provide an example if needed)
- you can even avoid the captcha to need sessions (by creating some tmp files yourself and deleting "old" ones via some cron). This is of course not a real recommendation, but might be an improvement.

liedekef:
A simple nonce example ($config contains your global config settings and a "key" which is just a string of random characters that can be initialized upon install)

```
function my_hash( $data ) {
        global $config;
        return hash_hmac( 'md5', $data, $config['key']);
}

$date = new DateTime();
$current_ts = $date->getTimestamp();
$hash = my_hash($current_ts);
```
Now you can give $hash and $current_ts as parameters to an url (or a POST in a form), and then when someone clicks on the link, you can check the value of $current_ts (by rehashing it and comparing against the hash that was also in the url). Once the check is done and the timestamp is valid, it is just a matter of coding to check the difference between that timestamp and the current time.
Example url (simplified): https://../?ts=$current_ts&nonce=$hash
Example verification ($config['form_submit_time'] contains the number of seconds a nonce is considered valid):
```
        $ts = intval($_GET['ts']);
        $calculated_hash = belnet_hash($ts);
        $date = new DateTime();
        $current_ts = $date->getTimestamp();
        if ($calculated_hash != $_GET['nonce'] || $current_ts-$ts>$config['form_submit_time']) {
                echo "invalid"; return false;
        }
```
For a POST, you give in the form the hash and ts as hidden params and then check in the post-action using the code above (but replacing $_GET by $_POST).

PR #612, opened 2021-12-18
fix apache / bullseye

### LINKED ISSUE #611: docker: image build fails and/or ldap config missing
Hi,
I'm having a hard time building the image with the default DOCKERFILE. 
It errors out when trying to install apache2 from the buster backports repo:
`E: Packages were downgraded and -y was used without --allow-downgrades.`

The strange thing is, when allowing downgrades, the final image is lacking the /etc/ldap directory altogether.
Consequently all binds to AD via ldaps fail with "Cannot access LDAP directory":
`[Sat Dec 18 10:37:06.422253 2021] [php7:warn] [pid 19] [client 10.0.1.3:60684] PHP Warning:  ldap_bind(): Unable to bind to server: Can't contact LDAP server in /var/www/html/sendtoken.php on line 89`

Please let me know, if more details are needed. Any help is highly appreciated.

faust64:
Should fix #611

### DIFF
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -11,7 +11,7 @@ RUN buildDeps=" \
         libonig-dev \
     " \
     runtimeDeps=" \
-	apache2/buster-backports \
+	apache2 \
         curl \
         libicu-dev \
         libldap2-dev \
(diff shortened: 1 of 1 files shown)


coudot:
Thanks @faust64

Issue #608, opened 2021-12-09
objectClass: ldapPublicKey should be added when needed


roumano:
Hi,

To add a ssh-key, the object class `ldapPublicKey` should added on the account,  
Otherwise the SelfServicePassword will get a `error 65` when trying to add the attribut ssh PublicKey on the Ldap directory.

Can you add a test if the object class is exist and if not add it before adding the ssh key ? 

Error without the objecClass : 
```
#!ERROR [LDAP: error code 65 - attribute 'sshPublicKey' not allowed]
dn: uid=dsmith,ou=People,dc=embl,dc=fr
changetype: modify
add: sshPublicKey
sshPublicKey:: ZGZzIGRmcyBkZxxxxxxx
```

Should be change to something like this :  
```
dn: uid=dsmith,ou=People,dc=embl,dc=fr
changetype: modify
add: sshPublicKey
sshPublicKey:: ZGZzIGRmcyBkZxxxxxxx
-
add: objectClass
objectClass: ldapPublicKey
-
```

Error on the apache log : 
```
PHP Warning:  ldap_mod_replace(): Modify: Object class violation in /var/www/password/lib/functions.inc.php on line 531, referer: https://XXXX/index.php?action=changesshkey
LDAP - Modify sshPublicKey error 65 (Object class violation), referer: https://XXXX/index.php?action=changesshkey
```

doc-slice:
I also need this extension and have programmed it out. How can I make this available?

coudot:
Saw your contribution, I made some comments in it

PR #618, opened 2022-01-05
add required objectClass for sshkey attribute, if needed

doc-slice:
Added a configuration parameter to be able to specify a required LDAP ObjectClass for the SSHKEY attribute. If the parameter is set, the system checks whether the specified ObjectClass is entered before changing the SSHKEY attribute. If not, it is added to the user. Issue #608


coudot (changes requested):
See https://github.com/ltb-project/self-service-password/pull/618#issuecomment-1005655834

coudot:
New code in #619

PR #619, opened 2022-01-05
Issue 608

doc-slice:
Issue #608 
- rename configuration attribute
- aggregate in same function, add/modify objectClass together with sshkey attribute in one operation
- translation not necessary, because no extra messages required
- documentation added

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -174,8 +174,8 @@ $change_sshkey = false;
 # What attribute should be changed by the changesshkey action?
 $change_sshkey_attribute = "sshPublicKey";
 
-# which objectClass is required for that attribute
-$required_sshkey_objectclass = "";
+# What objectClass is required for that attribute?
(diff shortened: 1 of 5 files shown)


conf/config.inc.php
```diff
@@ -174,6 +174,9 @@
 # What attribute should be changed by the changesshkey action?
 $change_sshkey_attribute = "sshPublicKey";
 
+# What objectClass is required for that attribute?
+$change_sshkey_objectClass = "";
```
coudot, line 178: I think we should set the default value to ldapPublicKey, as this is the standard objectClass.

The only impact would be for people using the attribute with another object class, but this should be seldom.

coudot (changes requested):

doc-slice:
default value is now set to ldapPublicKey, documentation adapted

coudot (approved):

coudot:
Thanks a lot!

PR #634, opened 2022-02-04
Spec file clean up

xavierba:
Hi,

I had this taking dust on a shelf, I guess it's better to have this upstream rather than in a fork :-)

Regards,
Xavier

### DIFF
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -21,7 +21,7 @@ RUN buildDeps="${BUILDDEP}" \
 	libgd3 \
 	libjpeg62-turbo \
 	libpng16-16 \
-	libwebp6 \
+	libwebp7 \
         libzip-dev \
 	locales \
 	locales-all \


coudot:
Hello,

thanks for this PR.

Some questions:
* Conf is moved to /etc/self-service-password, right? How will this work when updating the package?
* We advise to create a local configuration (config.inc.local.php), is this compatible with your packaging guidelines?

coudot:
Hello @xavierba 

did you saw my previous questions?

There are a lot of changes in your PR and I fear that it will break updates.

xavierba:
hi @coudot
Sorry for the very late answer.
I just rebased on top of master.

About the conf file, moving it from the older locations to the new one is taken care of in the %post section.
There's no problem with having a config.inc.local.php file as long as it is in the same directory as config.inc.php.
Probably the config.inc.local.php file should be moved at the same time as the main config.inc.php. I'll add another commit to do that.

xavierba:
The idea is to possibly include it in Fedora/EPEL and maintain it there. Both spec can indeed be kept in sync.  At least, cleaning up the upstream spec cannot hurt.

coudot:
My concern  is about users installing from our packages and knowing where the files are. It will impact all their documentations.

PR #642, opened 2022-03-07
Add some cosmetic css properties to sshkey textarea

spike77453:
This adds some cosmetic properties to the `sshkey` textarea. In particular:

- `resize: vertical;`
  This prevents a user from (accidentally) horizontally resizing the textarea which looks rather out of place:

Screenshot_20220307_120353

- `font-family: monospace;`
  A monospaced font seems more fitting for an ssh key

- `word-break: break-all;`
 This prevents a line break on a forward slash:

Screenshot_20220307_121829

vs.

Screenshot_20220307_121819

- `min-height: 10em;`
  This ensures the textarea is tall enough so a 2048 bit RSA public key (which is hopefully the smallest sized used these days) can be pasted without a scroll bar appearing and automatically scrolling past the beginning.  
  Otherwise one might wonder if the correct thing was pasted since it's not immediately obvious:
  
Screenshot_20220307_122254

Maybe it's even a good idea to make this tall enough for a 3072 bit RSA key since that's the default for ssh-keygen or 4096 bits.

### DIFF
--- a/htdocs/css/self-service-password.css
+++ b/htdocs/css/self-service-password.css
@@ -56,6 +56,13 @@ div#footer {
   width:100%;
 }
 
+textarea#sshkey {
+  font-family: monospace;
+  word-break: break-all;
+  resize: vertical;
+  min-height: 10em;
+}
+
 @media print {
 
   html, body {


coudot:
Thanks for the detailed report and the patch!

PR #655, opened 2022-05-23
AD doesnt accept TLS

vaslari:
Just adding some lines to config_ldap.rst
Specifying the TLS false for AD
It was a bit difficult for me to get it to work with AD because I didn't know it worked without TLS

### DIFF
--- a/docs/config_ldap.rst
+++ b/docs/config_ldap.rst
@@ -164,14 +164,13 @@ field:
 .. code:: php
 
    $ad_mode = true;
-   $ldap_starttls = false;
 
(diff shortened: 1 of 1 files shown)


coudot:
Hello @vaslari 

Seems AD is compatible with StartTLS : https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/8e73932f-70cf-46d6-88b1-8d9f86235e81

So the documentation should not force to disable StartTLS. As for other LDAP servers, the administrator should choose between LDAPS or LDAP+StartTLS

vaslari:
Understood, my bad.
Will check my installation then.
Thanks!

PR #661, opened 2022-06-19
Update bootstrap to v3.4.1

bohze:


coudot:
Thanks!

Will see if we work on a major upgrade of bootstrap for next version

Issue #658, opened 2022-05-30
Allow more than one mobile attribute


coudot:
Same as #494 but for mobile number: we should be able to check several attributes and get the first one which contains a value.

artlog:
i am working on it.

coudot:
Thanks @artlog !

PR #673, opened 2022-07-04
feat(sms): Allow more than one mobile attribute #658

artlog:
- replace $sms_attribute string by $sms_attributes array of string
  - update default configuration with attribute list
    mobile pager ipPhone homephone
  - update documentation
- provide a new function ltb_filter_attributes($ldap, $entry, $attributes)
  - ltd_ prefix to clarify what componenr provides this feature
    ie this is not in php_ldap extension
  - for future use to factorize this for existing mail feature
    - new class LtbAttributeValue
      keep track of attribute name and value binding
- update existing documentation for smssmailto
  - clarify {sms_attribute} usage
    this is not attribute name of configuration but actual gathered sms number

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -277,10 +277,10 @@ $use_sms = true;
 $sms_method = "mail";
 $sms_api_lib = "lib/smsapi.inc.php";
 # GSM number attribute
-$sms_attribute = "mobile";
+$sms_attributes = array( "mobile", "pager", "ipPhone", "homephone" );
 # Partially hide number
 $sms_partially_hide_number = true;
-# Send SMS mail to address
+# Send SMS mail to address. {sms_attribute} will be replaced by real sms number
 $smsmailto = "{sms_attribute}@service.provider.com";
 # Subject when sending email to SMTP to SMS provider
 $smsmail_subject = "Provider code";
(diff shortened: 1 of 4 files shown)


lib/functions.inc.php
```diff
@@ -833,3 +833,36 @@ function allowed_rate($login,$ip_addr,$rrl_config) {
     }
     return true;
 }
+
+class LtbAttributeValue {
```
coudot, line 837: Maybe it would be better to create a separate file for this class, to lighten the current file

coudot (changes requested):

lib/functions.inc.php
artlog: what filename do you suggest ?
should it include  ltb_ldap_get_first_available_value as well ?

coudot: I propose: LtbAttributeValue_class.php

For ltb_ldap_get_first_available_value, good question. I'm not a object oriented programming expert, but I think the class should only contains methods related to the object. So with this rule, the function should not be in the class. What do you think?

artlog: Having all within same file is better ihmo, and it can then be included only if needed.
To respect OOP then it can be made static method of LtbAttributeValue so
 ltb_ldap_get_first_available_value,  would be : LtbAttributeValue.ldap_get_first_value($ldap, $entry, $attributes)
and use a __construct to be able to call new LtbAttributeValue($attribute,$value) instead of getLtbAttributeValue

coudot: Seems ok for me

artlog:
new commit provided

coudot (approved):

PR #680, opened 2022-07-18
captcha use dedicated session cookie fix #602

### LINKED ISSUE #602: Errors when Using Captchas with Password Reset Emails

artlog:
- handle whole captcha statefull phrase read within function global_captcha_check
- there is a dedicated cookie named captcha used to find phrase corresponding
  to displayed image from server php session
- make sure to close session that will be recreated for token or smstoken purposes
  since it is not possible to keep multiple sessions open

TODO :
- fail captcha earlier without any additonnal checking.
- use other means to persist data on server side than using session.
- don't use cookie at all

### DIFF
--- a/htdocs/captcha.php
+++ b/htdocs/captcha.php
@@ -21,10 +21,19 @@
 
 require_once("../lib/vendor/autoload.php");
 use Gregwar\Captcha\CaptchaBuilder;
+
+# cookie for captcha session
+ini_set("session.use_cookies",1);
+ini_set("session.use_only_cookies",1);
+session_name("captcha");
 session_start();
 
 $captcha = new CaptchaBuilder;
 $_SESSION['phrase'] = $captcha->getPhrase();
+
+# session is stored and closed now, used only for captcha
+session_write_close();
+
 header('Content-Type: image/jpeg');
 $captcha
     ->build()
--- a/htdocs/change.php
+++ b/htdocs/change.php
@@ -32,17 +32,12 @@ $login = $presetLogin;
 $confirmpassword = "";
 $newpassword = "";
 $oldpassword = "";
-$captchaphrase = "";
 $ldap = "";
 $userdn = "";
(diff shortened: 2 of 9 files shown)


coudot:
Hello, tested it and it works. We indeed introduce cookies but this should not be blocking.

Would you like to keep working on this issue or are you confident enough to merge it?

artlog:
i'm confident enough that it can be merged, reworking it will be a longer term plan.

PR #684, opened 2022-07-20
Add rate limit checking for any password change request include fix #654

### LINKED ISSUE #654: Rate limit for Password reset
How can i develop a rate limit for a password reset function?

artlog:
- change password, change sshkey and reset by question are now protected by rate limit
- fix minor error in allowed_rate when ssp_rrl_ips.json can't be created


lib/functions.inc.php
```diff
@@ -884,6 +884,7 @@ function allowed_rate($login,$ip_addr,$rrl_config) {
             return false;
         }
     }
+
```
coudot, line 887:
```suggestion

```

PR #688, opened 2022-07-25
Fix password check ldap

artlog:


rest/v1/checkpassword.php
```diff
@@ -18,13 +19,14 @@
...
-    $ret = check_password_strength($newpassword, $oldpassword, $pwd_policy_config, $login, $entry);
+    $entry_array=array();
+    $ret = check_password_strength($newpassword, $oldpassword, $pwd_policy_config, $login, $entry_array);
     $result['error'] = 0;
-    $result['result'] = $ret;
-    $result['message'] = $messages[$ret];
 } else {
     $result['error'] = 1;
-    $result['message'] = "newpassword required";
+    $ret = "newpasswordrequired";
```
coudot, line 27: The message content should not be changed
```suggestion
    $ret = "newpassword required";
```

PR #699, opened 2022-08-23
Check parameters before calling hash_equals

coudot:
We check that parameters are strings before calling hash_equals, and avoid PHP fatal error


coudot:
@artlog can you check my code?

PR #718, opened 2022-09-18
Update config_apache.rst

souhaib22:
update root localisation


coudot:
There is a lot more places in this configuration file where `local` should be replaced by `share`

coudot:
Thanks!

souhaib22:
updated all /usr/local by /usr/share

PR #721, opened 2022-09-27
captcha in sendsms check when needed only fix #720

### LINKED ISSUE #720: SMS code won't send when captcha is enabled
After upgrading to version 1.5.1, I've found I can't reset the password using the SMS method when the captcha is enabled.

With the captcha enabled, when you press send on the "Check that user information are correct and press Send to get SMS token" page, it redirects you back to the original SMS page claiming "the captcha is required" with no code sent.

When the captcha is disabled, the code is sent and the user is redirected to the SMS token page after clicking send as expected.

No errors are written to the logs file when this happens.

This is on FreeBSD using nginx with the sms_method set to mail. Tested with both PHP74 and PHP80.

artlog:
- revert to previous behavior of sendsms to place captcha check.
  - sendsms is called mutliple times during sms sending process one the first need captcha check.

- was a regression of #680

### DIFF
--- a/htdocs/sendsms.php
+++ b/htdocs/sendsms.php
@@ -38,15 +38,7 @@ $token = "";
 $sessiontoken = "";
 $attempts = 0;
 
-#==============================================================================
-# Check captcha
-#==============================================================================
-if ($use_captcha) { $result = global_captcha_check();}
-
-if (! ($result === "") ) {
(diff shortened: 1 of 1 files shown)


coudot (approved):

coudot:
Thanks for the fix

PR #722, opened 2022-10-03
Use gpg instead of apt-key during for deb install

ZephOne:
apt-key is deprecated (see apt-key(8)), use of a gpg keyring should be used.


coudot:
Thanks!

PR #723, opened 2022-10-06
Update gpg install command

wwuck:
The default location for keyrings on Debian-based systems is `/usr/share/keyrings`. This directory already exists by default and should not need to be created.

This change also runs `sudo tee` so that `wget` and `gpg` are not run by root.

https://blog.cloudflare.com/dont-use-apt-key/

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -39,19 +39,13 @@ Configure the repository:
 
 .. code-block:: ini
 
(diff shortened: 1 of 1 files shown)


coudot:
> The default location for keyrings on Debian-based systems is /usr/share/keyrings. This directory already exists by default and should not need to be created.

Ok

> This change also runs sudo tee so that wget and gpg are not run by root.

This may not be really useful as all other commands are run by root

PR #754, opened 2023-02-13
Added comment/note over $custom_css

stevleibelt:
This is a small change.

I've added a small comment over `$custom_css`. The rest of the config is well documented, either by values or by comment. This change is there to improve the overall quality of the config to ease up the usage.

Thanks for the project.

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -325,6 +325,7 @@ $logo = "images/ltb-logo.png";
 # Background image
 $background_image = "images/unsplash-space.jpeg";
 
+# Path is relative to htdocs/html and should start with "css"
 $custom_css = "";
 $display_footer = true;
 
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -325,7 +325,7 @@ $logo = "images/ltb-logo.png";
 # Background image
 $background_image = "images/unsplash-space.jpeg";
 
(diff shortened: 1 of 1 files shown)


coudot:
Hello @stevleibelt 

thanks for this proposition. Not all parameters are documented in the config file, but they should all be listed in official documentation. For custom CSS: https://self-service-password.readthedocs.io/en/latest/config_general.html#custom-css

The config file should not be a repetition of the documentation. But if you think this would be better to have a comment for each attribute, we can discuss about it. Regarding your proposition, it does not seems clear enough, I would have written:
```php
# Path is relative to htdocs/html and the custom CSS file should be created in css/ directory. For example: "css/sample.css"

stevleibelt:
Hello @coudot 

thanks for your remark. I've updated the commend to your advice.

I must admit, that I prefer it, when the code (configuration file) is speaking for itself. I know there are different attitudes on this topic. But having the same content as comment in the code an in the documentation is fine for me. I treat the documentation as a place where you can write more expressive sentence for a topic. The more you have to write, the more the documentation expresses that this topic is not trivial.
This method would/should result in almost identical documentation and comment section when it comes to configuration files.

I prefer projects where I don't have to consult the documentation page when I am reading the configuration file :-).

Best regards,
Stev

coudot:
Thanks!

Issue #755, opened 2023-02-17
Host header poisoning


downgrad3:
Hello guys,
The password reset feature relying on mails is prone to an Host-header poisoning.
This allow an attacker to steal his victim's account reset token leading to an account takeover

[In sendtoken.php 185->209 ](https://github.com/ltb-project/self-service-password/blob/12617c61480ac5c4382de7581094ec1c2f235540/htdocs/sendtoken.php
) your product's relying on the Host header of the password-reset request to craft the password-reset mail. 
This user's input should be untrusted.

As an attacker, let's target the account "test.reset"
A password-reset request is performed on his account. In the request the HTTP Host Header value is tampered to set a arbitrary value (a server under the attacker's control here)

```
POST /?action=sendtoken HTTP/1.1
Host: **111.111.111.111**
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:102.0) Gecko/20100101 Firefox/102.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate
Content-Type: application/x-www-form-urlencoded
Content-Length: 16
Origin: https://portal-lab.ngp.infra
Referer: https://portal-lab.ngp.infra/?action=sendtoken
Upgrade-Insecure-Requests: 1
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: same-origin
Sec-Fetch-User: ?1
Te: trailers
Connection: close

login=test.reset
```

As the application rely on this value to craft the mail link. The victim recieve a mail with the following link

![2023-02-17_16h47_52](https://user-images.githubusercontent.com/11787001/219700829-0be81246-38e2-4216-adb6-2362c4de26ec.png)

If the victim click on the link, his reset token is sent to the attacker's server. 
The attacker can then use this token to perform account takeover

![image](https://user-images.githubusercontent.com/11787001/219702686-969bdab1-fd82-4542-abbc-9890a81e2d92.png)

coudot:
Note that this is only possible if SSP is not installed inside a virtual host.

The quick workaround is to configure `$reset_url`.

See also discussion in #756

PR #756, opened 2023-02-17
Checking server_name is known/trusted

faust64:
ensure the SERVER_NAME we use generating password reset links would point to a known FQDN
fix a couple missing locale strings
and ... why did we have html headers in ca.inc.php?!

see: https://github.com/ltb-project/self-service-password/issues/755

while I'm tempted to answer this is some webserver configuration issue: to be fair, our docs/samples would set self-service-password as a default vhost.
not something with kubernetes, or behind a reverse proxy.
still could affect "simple" installations.

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -390,6 +390,10 @@ $obscure_failure_messages = array("mailnomatch");
 # The name of an HTTP Header that may hold a reference to an extra config file to include.
 #$header_name_extra_config="SSP-Extra-Config";
 
+# Restricting valid server names, issuing SMS or Mail reset password links.
+# default to unset: would accept any value / legacy behavior
+#$valid_server_names = array();
+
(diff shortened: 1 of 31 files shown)


faust64:
TODO. some docs, most likely.
Not sure where to put this ... config_general, most likely ...

coudot:
Maybe the solution is to avoid constructing the reset URL dynamically? We already have the `$reset_url` parameter which is commented by default.

I don't see the advantage of adding a new parameter for allowed hosts.

faust64:
Maybe we should document with vhosts ... as you answered to the issue.. This is the obvious solution...
And yet, we give out samples like this: https://self-service-password.readthedocs.io/en/latest/config_nginx.html

I'm fine with the use-vhost solution. It's better than adding configs. But we can't be giving out samples contradicting this, then.

PR #757, opened 2023-02-21
Added JS to prevent multiple submits

stevleibelt:
I've added a little bit of javascript that disables the submit button once it is clicked.

The reason for that is to prevent users send the request form multiple times by clicking the submit button multiple times. If done so, this could result into an "false positive" error message that the password change was not successful.

### DIFF
--- a/templates/change.tpl
+++ b/templates/change.tpl
@@ -37,7 +37,7 @@
         {include file="policy.tpl"}
     {/if}
     <div class="alert alert-info">
-    <form action="#" method="post" class="form-horizontal">
+    <form action="#" method="post" class="form-horizontal" onsubmit="document.getElementById('submit-button').setAttribute('disabled', '')">
         <div class="form-group">
             <label for="login" class="col-sm-4 control-label">{$msg_login}</label>
             <div class="col-sm-8">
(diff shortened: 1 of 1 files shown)


coudot:
Not a bad idea, but I would prefer not to put JS code in template, we should use htdocs/js/self-service-password.js

stevleibelt:
@coudot I will give the adaption a try. Sadly, my JavaScript-Skills are incredible low. Just give me a bit of time to adapt my pull request.

Best regards,
Stev

coudot:
https://gomakethings.com/dont-disable-buttons/

stevleibelt:
For both cases and in the role of the user, the system should inform the user that the form has been sent already and that the webpage is waiting for a response.

I personally like a disabled button more than a button I can click on but no action is done since I have a instant feedback from the system that this button does not trigger any action right now. But really, that is just my gut feeling.

Since you are the lead here @coudot, what would you prefer?
Since the year is coming to an end, I would like to finish this long going open task with an happy merge before 2024 :-).

Best regards

coudot:
I'm not an expert but it seem that using `event.preventDefault();` is not hard to do. It would indeed be a good thing to change CSS class or if you want, disable it.

You can add this in htdocs/js/self-service-password.js and make it available for all SSP forms

davidcoutadeur:
I have created a dedicated issue here: #844

See my PR #845 for fixing the issue with a preventDefault function. Tested on my desktop successfully

davidcoutadeur:
If ok, we can close this PR

stevleibelt:
@davidcoutadeur Thanks for your way better fix!

I will close this PR in favor of your [844](https://github.com/ltb-project/self-service-password/issues/844).

PR #770, opened 2023-05-09
Clarify the message about passwords being found in HIPB

gnyman:
The old password was a bit vague and I believe confusing. I have tried to improve the messaging.

I am aware the inclusion of FBI might be an americanism but I included it still as FBI does feed data into Have I Been Pwned (HIPB) and my thinking is that that it is well known and respected enough to gives some authority to the messaging.

I did consider "law enforcement agencies" also considered that too vague. When updating the other translations, other more local wording might be better.


coudot:
I'm not sure this should be the default message, indeed the reference to FBI seems not useful here.

Note that you can easily customize the message by adding a lang file in conf/ folder.

gnyman:
Thanks for your feedback. I know it's easy to change, but I think in the spirit of "secure by default" one should try to make the default message as useful as possible. As I think it's unlikely most users will spend the time to change any of the defaults.

What do you think about the change if we would change "FBI" to something more neutral like "law enforcement agencies" ?

PR #799, opened 2023-09-06
Update phpunit

coudot:
* Update tests to recent phpunit versions
* Add a workflow in github

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,12 @@
+name: CI
+
+on: [push]
+
+jobs:
+  build-test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - uses: php-actions/composer@v6
+    - uses: php-actions/phpunit@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,5 +8,4 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - uses: php-actions/composer@v6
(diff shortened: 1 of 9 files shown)


markus-96:
It seams to me that you are missing some configuration. I got it to work with these additions:
In ci.yml, replace `uses: php-actions/phpunit@v3` with the following:
```
        - name: PHPUnit Tests for php7.4
        uses: php-actions/phpunit@v3
        env:
          TEST_NAME: Scarlett
        with:
          bootstrap: lib/vendor/autoload.php
          configuration: tests/phpunit.xml
          version: 5.7.25
          php_version: 7.4
          args: --coverage-text
```
This will test against php7.4, for other versions of php you can add additional blocks. 

The phpunit.xml can look like this (working for me):
```
<?xml version="1.0"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" colors="true" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd">
	<coverage>
		<include>
			<directory suffix=".php">../../src</directory>
		</include>
		<report>
			<html outputDirectory="./_coverage" lowUpperBound="35" highLowerBound="70"/>
			<text outputFile="php://stdout" showUncoveredFiles="true"/>
		</report>
	</coverage>
	<testsuites>
		<testsuite name="main">
			<directory suffix="Test.php">.</directory>
		</testsuite>
	</testsuites>
	<logging/>
</phpunit>
```

The things about coverage tests can be removed I think..

markus-96:
> Thanks @markus-96, it is now working!

Happy to hear that!

Issue #803, opened 2023-09-20
German Translation of some parts of the code


piang0:
### Discussed in https://github.com/ltb-project/self-service-password/discussions/802

<div type='discussions-op-text'>

<sup>Originally posted by **piang0** September 20, 2023</sup>
Hello together. First thank you a lot for this amazing project. :) 
I found this a short time ago and have already configured it successfully. What I noticed is that there are a few small errors in the German translation or a few places are not translated at all. 
For example, I found this and translated into German: 

Original: 
`$messages['tokensent_ifexists'] = "If the account exists, a confirmation email has been sent to the associated email address";`

Translation: 
`$messages['tokensent_ifexists'] = "Wenn das Konto existiert, wurde eine Bestätigungs-E-Mail an die zugehörige E-Mail-Adresse gesendet";`

Is there any way to help with the translation or is there already a process for this ? 

Thank you and best regards
Max</div>

coudot:
Sure, you can submit a Pull request with your modifications

PR #804, opened 2023-09-20
translated the line $messages['tokensent_ifexists'] into german 

piang0:
As described in my Issue #803, I have translated the corresponding line into German. :)


coudot:
Thanks!

PR #807, opened 2023-10-03
explicitly use docker image with Debian 11

sraisl:
apparently the Debian Version in the php:8.1-apache image has been changed to Debian 12. Some runtime dependencies, like libwebp6 are not available in  Debian 12. Therefore the php image tag should be changed to use Debian 11 explicitly.


coudot:
Hello @sraisl 

this is a correct workaround, but I would prefer to switch to Debian 12 and use libwep7

davidcoutadeur:
This issue has been fixed in: Dockerfile fails to build #813 #814

Issue #288, opened 2019-04-18
Page to set mail for use by "reset by token" feature


mohaja:
I would like to request a page where user can set/change email address for use later in 'reset by token' page. Interface will be much like the change password with extra field for email address, and captha to prevent spamming. Anyway, great app and very useful in my school. Thanks

coudot:
Hello,

we will see how to do it.

For the moment, we prefer to use FusionDirectory (https://www.fusiondirectory.org/) to let users change their informations.

PR #808, opened 2023-10-03
First implementation of a page to set mail and phone attributes

coudot:
The goal is to let users modify their mail or phone in LDAP directory

Linked to #288


coudot:
Documentation needed before accepting the PR

PR #839, opened 2024-02-19
Update it.inc.php

iotaka:
Fixed wrong accent the verb "to be" is with "è" not "é" in Italian. Changed some sentences.


coudot:
Thanks!

PR #845, opened 2024-03-06
prevent multiple form submits (#844)

### LINKED ISSUE #844: Added JS to prevent multiple submits
Proposition taken from @stevleibelt : https://github.com/ltb-project/self-service-password/pull/757

Original proposal:

I've added a little bit of javascript that disables the submit button once it is clicked.

The reason for that is to prevent users send the request form multiple times by clicking the submit button multiple times. If done so, this could result into an "false positive" error message that the password change was not successful.

davidcoutadeur:
Closes #844


coudot:
Thanks!

Issue #165, opened 2017-11-23
Multiple SMS tokens could be requested if an attacker finds one username, leading to big expenses


Emi94:
Hi,
Reset by SMS requires only the username to send a token. It is not very difficult for an attacker to find one username and use it to make multiple requests to reset by SMS, leading to possibly big expenses to the SMS provider. I recommend it to ask the user for the phone number too and check that it matches with the one from the LDAP, same as it's done for reset by email. In advance it would be good to have the possibility to set a period of time that needs to pass before the same user can request another reset by SMS.

coudot:
@Worteks is looking at it (thanks @armfem)

PR #851, opened 2024-03-08
Securisation of reset by SMS token feature

armfem:
Solving the Issue #165 multiple sms tokens. Changing default behaviour for sms sending, now it asks for telephone number by default.


htdocs/sendsms.php
```diff
@@ -96,32 +111,46 @@
...
+        if (!$sms_use_ldap) {
+            # TODO: Add possibility to loop over different phone numbers.
+            #foreach ($mailValues as $mailValue) {
+            if (strcasecmp($sms, $phone) == 0) {
+                $match = true;
+                $encrypted_sms_login = encrypt("$sms:$login", $keyphrase);
+                $result = "sendsms";
+            }
+            if (!$match){
+                $result = $obscure_usernotfound_sendtoken ? "tokensent_ifexists" : "smsnomatch";
```
coudot, line 141: If `$obscure_usernotfound_sendtoken` we should always display the screen requesting the SMS code, to avoid indicating to an attacker that the user is not found. We should only display an error if this option is false.

coudot (changes requested):

htdocs/sendsms.php
```diff
@@ -96,32 +111,46 @@
...
-    if ($sms) {
-        $encrypted_sms_login = encrypt("$sms:$login", $keyphrase);
-        $smsdisplay = $sms;
-        if ( $sms_partially_hide_number ) {
-            $smsdisplay = substr_replace($sms, '****', 4 , 4);
+    if ($sms){
+        if (!$sms_use_ldap) {
+            # TODO: Add possibility to loop over different phone numbers.
+            #foreach ($mailValues as $mailValue) {
+            if (strcasecmp($sms, $phone) == 0) {
```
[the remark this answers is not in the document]
armfem: Added same filter for `$phone`

htdocs/sendsms.php
```diff
@@ -96,32 +111,46 @@
...
     [$result, $sms, $displayname] = get_mobile_and_displayname($login);
-
-    if ($sms) {
-        $encrypted_sms_login = encrypt("$sms:$login", $keyphrase);
-        $smsdisplay = $sms;
-        if ( $sms_partially_hide_number ) {
-            $smsdisplay = substr_replace($sms, '****', 4 , 4);
+    if ($sms){
+        if (!$sms_use_ldap) {
+            # TODO: Add possibility to loop over different phone numbers.
```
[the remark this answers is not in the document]
armfem: Deleted the comment.

htdocs/sendsms.php
armfem: This is the new default behaviour, and a new option has been added to the configuration to change into displaying all the information.

armfem: The new variable is `$obscure_notfound_sendsms = true;`

htdocs/sendsms.php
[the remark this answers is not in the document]
coudot: Fixed a typo in the functions, seems ok now

htdocs/sendsms.php
coudot: I did some tests, what I see for now:
* the message `smssent_ifexists` is defined in lang files but never used in templates
* the hidden field token is empty if user and/or phone is incorrect, which allows the attacker to know if the search in LDAP fails. We need to generate a token in all cases

armfem: I've corrected those two problems. The branch have been updated.

coudot (approved):

PR #865, opened 2024-03-13
ability to change custom password fields, developped by @markus-96 (#864)

### LINKED ISSUE #864: feature: ability to change custom password fields
This feature is proposed and developed by @markus-96

This issue is for discussing the details of this feature.

Here is the original discussion in PR: #751

```
I think I have written everything needed for this feature. Only the messages shown on the page have to be changed, and also the translations. I will try my best here. I will remove [in progress] when everything is finished and after some additional testing.

You will be able to:

    setup as many app-passwords as you want,
    set the name of the app and the corresponding attribute,
    setup additional password policies,
    setup who changes the password,
    the hash algorithm

You can use it also for updating user information, but I do not think this would be very user friendly.

Also, take care about the discussion regarding this feature: https://github.com/ltb-project/self-service-password/discussions/745
```

davidcoutadeur:
Closes #864


davidcoutadeur (commented):
@markus-96

Here is a bunch of new remarks / comments:

- basically, the code in `changecustompwdfield.php` is not working, there is much work to adapt it. For example, the checks on confirmpassword, newpassword and oldpassword are invalid, because you have replaced these fields by confirmcustompassword, newcustompassword and password. This is just an example, there are many other things to fix. Could you get the new branch `change-custom-password-fields` and fix these problems?

- there is a big duplication of code in `changecustompwdfield.php`. Basically, this file is a copy / paste of change.php. I think we should make much more factorization here. For now, I don't know exactly what is the best approach, so just ignore it for the moment.

- as you mentioned this before, some ppolicy checks are unusable, typically the pwd_no_reuse and pwd_diff_last_min_chars ones. The new ppolicy js code can be used for disabling these checks automatically. I will take this point in charge. The entropy of app-password could be displayed however. Why did you disable it?

- the code in `lib/functions.inc.php` can be factorized from https://github.com/ltb-project/ltb-ldap/pull/7 I'll give a look to this PR.

- check_password_strength and change_password signatures have changed, but there are still multiple places where these functions are called without the new arguments

davidcoutadeur:
Basically, the feature seems to work on my side.

I must do additional checks and especially give a look at the checkentropy modifications

davidcoutadeur:
Seems all good to me.

Many small fixes here:
- simplification of the checkentropy check called by js (can be called by any page)
- adding a new parameter: pwd_unique_across_custom_password_fields (see documentation)
- fill-in documentation
- simplification of default values
- bug fixes

I have tested extensively the feature. The only bug I encounter sometimes is that the field that is modified is not the custompwdfield, but the userPassword. I suspect there is maybe a wrong redirection to index.php at some point, but I can't figure out where is the problem. (and can't reproduce)

I write it here in case we reproduce one day the issue.

@markus-96 could you do a last test of the feature, and tell me if everything is ok for you? If so, I'll squash and rebase the pull-request. Thank you for your multiple reviews and tests during this long-standing issue.

htdocs/change.php
```diff
@@ -158,8 +158,8 @@
 #==============================================================================
 # Check password strength
 #==============================================================================
-if ( !$result ) {
-    $result = check_password_strength( $newpassword, $oldpassword, $pwd_policy_config, $login, $entry_array );
+if ( $result === "" ) {
```
markus-96, line 161: this has to be `if ( !$result ) {`.

Besides that, everything seams to be working as expected.

davidcoutadeur: Perfect, I have just fixed it in last commit. I am going to merge this

PR #871, opened 2024-03-18
ability to change custom password fields - working PR

markus-96:
Hello @davidcoutadeur, as you have proposed I deleted my fork and recreated it, with all branches included (sigh). For continuing working on it, I had to create this PR.

Original PR: #751
Current feature branche's PR: #865 
Issue ticket for the feature: #864


markus-96:
>     * basically, the code in `changecustompwdfield.php` is not working, there is much work to adapt it. For example, the checks on confirmpassword, newpassword and oldpassword are invalid, because you have replaced these fields by confirmcustompassword, newcustompassword and password. This is just an example, there are many other things to fix. Could you get the new branch `change-custom-password-fields` and fix these problems?

Is fixed, it was because of the check `if ( $result === "" )`, but the value of `$result` is set to `false` now in this line: `$result = $ldap_connection[1];`. Probably this was changed when introducing Ltb\Ldap.

Little sidemark: as I see it, in `change.php`, the password strength is currently not checked, because of this.

>     * as you mentioned this before, some ppolicy checks are unusable, typically the pwd_no_reuse and pwd_diff_last_min_chars ones. The new ppolicy js code can be used for disabling these checks automatically. I will take this point in charge. The entropy of app-password could be displayed however. Why did you disable it?

`pwd_no_reuse` is mandatory for this feature, see the changes in `check_password_strength`! As I already said, in my oppinion entropy checks are completely useless in this feature:

> For entropy checks:
> This can not be done between "app"-passwords and the user-password, because on the app-password side you need to know the old password, which you do not. So: While changing the "app"-password, you can check the entropy between "app"-password and user-password, but if you change the user-password, you can not check the entropy to the already set "app"-password. So this check is useless in this aspect.

markus-96:
@davidcoutadeur I totally misunderstood what the entropy checks are for. I thought it would indicate the level of difference between newpassword and oldpassword, sorry for that. Will delete the line disabling it.

markus-96:
edge case: you want to use $pwd_display_entropy for the custom field, but not for the main password. The action `"checkentropy"` has to be available, but may not be shown on the main password change page, but on the custom field page. So, there has to be a check: if `$pwd_display_entropy` is turned off for the main password, but turned on for a custom field, the action `"checkentropy"` still has to be available. This can, as I see it, only be done with a for loop.

Another option would be to introduce an additional config parameter, for example `$use_entropy`, for enabling/disabling this action. Then, you can decide for main and custom pwd fields independently with `$pwd_display_entropy` or `$change_custompwdfield[$i]['pwd_policy_config']['pwd_display_entropy']`.

I have implemented the for-loop for now, but `$use_entropy` seams much cleaner to me.

davidcoutadeur:
@markus-96 I have made a big refactoring of ltb-ldap library (no more global variables, new tests, new functions)

I have adapted the modifications in self-service-password in master branch. Could you check everything is ok for you and adapt your PR if necessary?

markus-96:
> @markus-96 I have made a big refactoring of ltb-ldap library (no more global variables, new tests, new functions)

As far as I can see this looks great! I may have something in the pipeline for managing the multiple bind strategies, but it is far from beeing ready. For example, as far as I can see it, you could make the ldap connection a property of the class Ldap and implement only the needed LDAP function that call the functions in \Ltb\PhpLdap, but with the property ldap instead of a variable. This would capsulate the rather delicate LDAP connection. I made something 5 days ago, but did not publish yet, but can for now be seen here: https://github.com/markus-96/ltb-ldap/tree/ldap-factorize

> I have adapted the modifications in self-service-password in master branch. Could you check everything is ok for you and adapt your PR if necessary?

This will have to wait until weekend, but shure!

markus-96:
@davidcoutadeur could you please sync ltb-project:change-custom-password-fields with main? (so that I can then update my branch)

davidcoutadeur:
> @davidcoutadeur could you please sync ltb-project:change-custom-password-fields with main? (so that I can then update my branch)

I have just rebased the change-custom-password-fields branch on top of master. It should be ok now

markus-96:
Will do some final testings with my instance of ds389, do not have anything else on my hands right now

markus-96:
The entropy checks give me some headaches. If `$pwd_display_entropy` is enabled, then `checkentropy.php` is loaded with require_once, but only then it is enabled. In `checkentropy.php`, there is an additional check whether `$pwd_display_entropy` is true. Why, since it will always be enabled? For now, it is working, but with some really clunky for loops, that would not be needed if there would be an option like `$use_entropy`..

Despite that, everything seams to work as expected.

davidcoutadeur:
I have just pushed your commits to the `change-custom-password-fields` branch, and made some small adaptations linked to the new ltb-ldap function calls.

Basically, the feature seems to work also on my side.
I must do additional checks and especially give a look at the checkentropy modifications

lib/functions.inc.php
```diff
@@ -194,6 +194,25 @@ function check_password_strength( $password, $oldpassword, $pwd_policy_config, $
         }
     }
 
+    # is same as a custom password?
```
davidcoutadeur, line 197: Hello,

I am doing a last big review for this feature.

@markus-96 This pwd_no_reuse check seems really strange to me.

You require that the new password for "customfield1" is different from:
- standard userPassword
- plus any configured "customfieldN"

This seems, at least, very specific, and confusing for the user. It is especially confusing for the admin because `pwd_no_reuse` does not have the same meaning as in the general configuration.

Also, I think there will be use cases when the password must be the same as other custom password field. With this option, you can't tune precisely what field must be different.

For the sake of simplification, my first idea would be to remove the feature, and only keep the simple logic of verifying only the field you are trying to update. Do you have a special use case for requiring this feature @markus-96 ? Could you describe the feature to check how we could implement it?

markus-96: > Hello,
> 
> You require that the new password for "customfield1" is different from:
> 
>     * standard userPassword
> 
>     * plus any configured "customfieldN"

(for clarification: it does not check any configured customfieldN, it only checks the one with $custompwdfieldN['pwd_policy_config']['pwd_no_reuse'] set to true)

> Do you have a special use case for requiring this feature @markus-96 ?

Well, the main focus for me was to have a separate password that is stored in sambaNTPassword, and I did not want it to be the same as the main userpassword, since it uses some rather insecure hashing method. If you change the password stored in sambaNTPassword, you can directly check if it is the same as the main userpassword. But if you change the main userpassword, you have to check if it is the same as the one that is represented by the hash stored in sambaNTpassword.

Setting the same password to the customfield as to the main userpassword, for me this sounded like reusing the password. This is why I used this setting key for this. But it is right, when turning this on, it should be checked if the new password is the same as the already set one.

So, I think it should be renamed to 'must_be_unique' or something like this.

These two considerations would result in a bit more code to be written. But this feature is mandatory for me, as I already said a few times, and without it I would it would be useless for me.

markus-96: > Other point: from time to time, the custompasswordfield is not modified in the LDAP, but the userPassword instead. I'll continue the investigation on this point.

Ok, this is really strange. Do you have any hint when this happens? I will also recheck everything that could be related to this, but in the time of writing this I have no idea how this could happen. Maybe one of these:

- when calling change_password, somehow $custom_pwd_field_mode is set to false or $custom_pwd_attribute is set to "userPassword"
- with ad_mode, maybe, somehow, something is not working as intended. But I can not check this since I do not have access to an AD instance, and ad_mode is always false here.

davidcoutadeur: > (for clarification: it does not check any configured customfieldN, it only checks the one with $custompwdfieldN['pwd_policy_config']['pwd_no_reuse'] set to true)

The problem here is that it checks all attributes having $custompwdfield[$n]['pwd_policy_config']['pwd_no_reuse'] == true

For example, if I have a configuration with 3 LDAP custom password attributes : "custom1", "custom2" and "custom3", when I am updating custom1 with password secret, ssp will check if this value is matching the ones in custom1, custom2 and custom3. What would be more logic IMO is to check only custom1 value.

davidcoutadeur: > So, I think it should be renamed to 'must_be_unique' or something like this. These two considerations would result in a bit more code to be written. But this feature is mandatory for me, as I already said a few times, and without it I would it would be useless for me.

Then I think it would be better to have a dedicated parameter, like, as you proposed: `unique_across_password_fields`

davidcoutadeur: If you are ok with the new parameter, I can work on this.

htdocs/checkentropy.php
```diff
@@ -66,5 +67,12 @@ function checkEntropyJSON($password_base64)
     $newpass = htmlspecialchars($_GET["password"]);
     checkEntropyJSON($newpass);
 }
+if(isset($change_custompwdfield)) {
+    foreach ($change_custompwdfield as $custompwdfield) {
+        if ( isset($custompwdfield['pwd_policy_config']) && isset($custompwdfield['pwd_policy_config']['pwd_display_entropy']) && $custompwdfield['pwd_policy_config']['pwd_display_entropy']) {
+            checkEntropyJSON($newpass);
+        }
+    }
+}
```
davidcoutadeur, line 76: Actually, this check is not correct, because as soon as one pwd_display_entropy has been enabled for any custompwdfield, the entropy check will be done.

Anyway, this is not a big deal, because I don't think the initial check: `if( isset($pwd_display_entropy) && $pwd_display_entropy == true )` is really necessary. This check is already done at the frontend side

I will simply remove it.

PR #873, opened 2024-03-18
Use a dedicated CSS class for ppolicy div

coudot:
DO not use `help` CSS class so that ppolicy js will only act on its div


davidcoutadeur:
Seems all fine to me

Issue #860, opened 2024-03-12
SMS sending option error when encryption is not activated


armfem:
According to the comments in the file `config.inc.php`, the use of sms needs `$crypt_tokens=true`.
This is the default value for `$crypt_tokens`. 

Nevertheless, if someone is to change this to `false`, the `sendsms.tpl` will only restart itself without any error for the user. There is an error shown in the logs.

Hence, the user would be able to access the `sendsms` template but it will never work.  
 
It will be better to force the application to have either both (encryption of token and sending sms) or none.

PR #874, opened 2024-03-19
Test crypt_tokens parameter if SMS feature is enabled

coudot:
Fix for #860


coudot:
@armfem could you review it?

PR #879, opened 2024-04-03
use new ltb-ldap v0.2 and adapt method signatures (#878)

### LINKED ISSUE #878: use new ltb-ldap v0.2
Especially, we need to adapt the new method signature to avoid global variables.

See https://github.com/ltb-project/ltb-ldap/issues/4

davidcoutadeur:
Closes #878

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "require": {
-        "ltb-project/ldap": "v0.1.0",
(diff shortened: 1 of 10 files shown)


davidcoutadeur:
Refactoring seems ready.

We will have to adapt the version of ltb-ldap (v0.2.0) in composer.json when released.

PR #880, opened 2024-04-04
Use ltb ldap functions in functions.inc.php

markus-96:
needs verification that everything works as expected, maybe a unit test using mockery


davidcoutadeur:
Done in #882

Thank you @markus-96 for your work here.

PR #890, opened 2024-04-18
Correction of max number of attempts for token sent by sms bug

### LINKED ISSUE #861: Bug in max_attempts for token in `sendsms`
When defining max number of attempts allowed for sms token, self-service password uses the convention to start counting at 0. 

Therefore, when max attempts is configured as 3, the user has actually 4 attempts.

armfem:
Added more explicit variable name and correcting number of token tries.

Closes #861

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /vendor/
 /htdocs/vendor/
+composer.lock
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -314,7 +314,7 @@ $sms_truncate_number_length = 10;
 # SMS token length
 $sms_token_length = 6;
 # Max attempts allowed for SMS token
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 /vendor/
 /htdocs/vendor/
+composer.lock
```
findlayfeng, line 3: If we start ignoring composer.lock, should we delete it?

findlayfeng (commented):
我认为配置相关的更新应当单独commint
---
I think configuration-related updates should be committed separately

.gitignore
davidcoutadeur: Good remark. Thanks for pointing at this issue. I have opened a dedicated issue for discussing about it: #929

Issue #912, opened 2024-06-12
PHP notices showing in the web ui


maichouni-mitek:
Hi,

When a user tries to reset their password, and puts in an erroneous old password, they get a PHP notice.
I believe this is not wanted behavior: it leads non-technical users to confusion, and it discloses information that they do not need (not a good security practice).
Screenshot to illustrate the issue: 
![image](https://github.com/ltb-project/self-service-password/assets/104677672/1a7bff02-4085-4f75-8c50-c41d8d63b119)
I think this is due to the fact that `error_reporting` is set after `change.php` is included.
I will be submitting a PR.

Thanks.

PR #913, opened 2024-06-12
fix: hide php notices from the end user in the web ui

maichouni-mitek:
- Closes https://github.com/ltb-project/self-service-password/issues/912.
- This change hides the PHP notices from the end user in the web ui, but still keeps them in the logs.
- `error_reporting` needs to be set before `change.php` is included.


coudot:
Thanks a lot!

PR #926, opened 2024-06-28
Add Arabic language support in docs

toky0:


coudot:
Hello, thanks for this proposition.

My question is about the flag, is it the correct one? I checked https://en.wikipedia.org/wiki/List_of_Arab_flags but did not found a common flag.

toky0:
Hi, Thanks for the quick review.

You're right, there's no common flag. We generally use one with the four colors white, green, red, and black like UAE, PS, or JO. I picked PS here.

coudot:
Could you rebase on master?

coudot:
Rebase done in #989

PR #932, opened 2024-07-02
修改了 Dockerfile 文件 | Modified Dockerfile 

findlayfeng:
* 修改了运行时依赖的处理方式，通过对 PHP插件的库连接情况自动处理保留的包
* PHP 的默认的配置文件通过link的方式链接到 production 文件
* 修改了 apache 的默认 root 文件夹和其他环境做统一
* 通过版本 环境变量SMARTY_VERSION 安装不同版本的 SMARTY
* 通过 Multi-stage builds 的方式构建镜像
* 删除了 .dockerignore 文件 通过 INSTALL_PATHS 变量的方式指定需要发布的路径，好处是添加了新的路径后需要显式的添加到发布列表，而不是默认发布，这个文件 5年没有修改了，显然的将不少不必要的文件发布到了docker镜像
---
* Modified the handling of runtime dependencies, automatically handling the retained packages by checking the library connection of the PHP plugin
* PHP's default configuration file is linked to the production file through link
* Modified the default root folder of apache to unify with other environments
* Install different versions of SMARTY through the version environment variable SMARTY_VERSION
* Build images through Multi-stage builds
* Deleted .dockerignore file The path to be published is specified through the INSTALL_PATHS variable. The advantage is that after adding a new path,
it needs to be explicitly added to the release list instead of being published by default. This file has not been modified for 5 years. Obviously, many unnecessary files are published to the docker image.

PR #947, opened 2024-07-24
946 improve docker image

### LINKED ISSUE #946: Improve docker image
This issue was proposed and developed by @findlayfeng 

- Modified the handling of runtime dependencies, automatically handling the retained packages by checking the library connection of the PHP plugin
- PHP's default configuration file is linked to the production file through link
- Modified the default root folder of apache to unify with other environments
- Install different versions of SMARTY through the version environment variable SMARTY_VERSION
- Build images through Multi-stage builds
- Deleted .dockerignore file The path to be published is specified through the INSTALL_PATHS variable. The advantage is that after adding a new path, it needs to be explicitly added to the release list instead of being published by default. This file has not been modified for 5 years. Obviously, many unnecessary files are published to the docker image.
- add an alpine image

davidcoutadeur:
Closes #946


davidcoutadeur:
This was originally proposed by @findlayfeng in [ #932](https://github.com/ltb-project/self-service-password/pull/932)

See the merge request above for the discussions

PR #949, opened 2024-07-24
fix resending mail requests (#401)

### LINKED ISSUE #401: Keep sending reset requests
On the page "Mail Confirmation has been sent" if i press F5 keeps sending the email over and over.

davidcoutadeur:
Closes #401

### DIFF
--- a/htdocs/sendtoken.php
+++ b/htdocs/sendtoken.php
@@ -32,6 +32,7 @@ $ldap = "";
 $userdn = "";
 $token = "";
 $usermail = "";
+$formtoken = "";
 
 if (!$mail_address_use_ldap) {
     if (isset($_POST["mail"]) and $_POST["mail"]) {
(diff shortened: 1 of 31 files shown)


htdocs/sendtoken.php
```diff
@@ -48,7 +49,26 @@
...
+}
+
+if (! isset($_POST["mail"]) and ! isset($_REQUEST["login"])) {
+
+    $result = "emptysendtokenform";
+
+    # Generate formtoken
+    ini_set("session.use_cookies",0);
+    ini_set("session.use_only_cookies",1);
+    ini_set("session.use_strict_mode",0);
```
coudot, line 63: We use `ini_set` in a lot of places in our code, and here in this page.

Maybe we should move them to index.php?

htdocs/sendtoken.php
```diff
@@ -162,14 +182,42 @@
...
+    }
+    session_write_close();
 }
 
 #==============================================================================
 # Send token by mail
 #==============================================================================
 if ( !$result ) {
 
+
```
coudot, line 220: Was it intentional to add this blank line?

htdocs/sendtoken.php
davidcoutadeur: Good point!

Maybe at some point we would need to have cookies sent for saving the session at client side. So, I am rather thinking to a dedicated function.

Also I am not quite satisfied by calling ini_set, whereas:
1. there are some warning in debug mode about misusage of ini_set function
2. it seems you can set those parameters in the session_start function (but my tests don't work for now)

htdocs/sendtoken.php
davidcoutadeur: No. I'll remove it

htdocs/sendtoken.php
coudot: Yes, using php session mechanism was an easy way to deal with tokens, but we may find a better solution. It will require a token backend on server side, which should be something easy to set up (files, in memory base, ...)

davidcoutadeur: Unfortunately, I don't see a simple and clean way for doing this. Indeed, in `lib/captcha/InternalCaptcha.php` we also use ini_set, but in this context we need cookies.

I think the best would be to find and use a php lib for managing server sessions. I am opening a dedicated issue for that. Not sure we can fix this for next 1.7.0 release though.

PR #950, opened 2024-07-26
Making it possible to docker build both on M1/M2/M3 Mac and in an x86 VM

maglub:
Simple fix to make it possible to build on Mac as well as on x86.


davidcoutadeur:
Thanks for your contribution. It will be included in 1.7.0 release.

maglub:
Thanks a lot!

Issue #850, opened 2024-03-08
Update to smarty4


davidcoutadeur:
Some time ago (see https://github.com/ltb-project/self-service-password/issues/773) we decided to remove smarty3 from direct dependencies, because

1/ it was broken on ubuntu (https://bugs.launchpad.net/ubuntu/+source/smarty3/+bug/1986916) and
2/ it was unavailable on some rh-like systems

I think it is interesting to now consider the upgrade to smarty4, for multiple reasons:
- smarty4 is released since 2021. It is now considered as the stable version. smarty3 is now only recommended for applications requiring PHP < 7.1 
- smarty4 seems well-packaged on multiple systems (tested and working on debian 12)
- we could simplify installation process by simply enabling smarty4 as a requirement in rpm/deb packages

coudot:
We also need to see if Smarty is a system dependency or a composer dependency

davidcoutadeur:
> We also need to see if Smarty is a system dependency or a composer dependency

smarty4 is a system dependency on debian.

@xavierba is working on a smarty4 release on rh-like systems

davidcoutadeur:
NOTE: When (if) updating requirement to smarty4, we must also list the new version in deb package (replacement of smarty path in config files in `debian/replace_cache_dirs` + Depends in `control` file +  ) and in RPM package (replacement of smarty path in config files + requirement of smarty package)

coudot:
php-Smarty 4 not yet available on EPEL/Fedora, but as we don't require explicitely v3, it will use Smarty 4 as soon it is available.

PR #969, opened 2024-09-02
Fix wrong smarty path

GloomyNAN:


davidcoutadeur:
Unfortunately, smarty 4 is not available in all distros. (especially red-hat ones)

See this issue: #850

You can still declare it manually in your config.inc.local.php file

PR #1030, opened 2025-01-08
Fix input-group on sendtoken (mail) page

ludovi-com:
Moved the hidden token input outside the Bootstrap input-group to ensure the correct display of the border-radius at the end of the text input


davidcoutadeur:
Closes #1036

PR #1051, opened 2025-03-19
Allow customized template

coudot:
This adds the `$custom_tpl_dir` option in configuration which can be used to create a customized template.

You can just override one or two templates files by copying them to the custom template directory and adapt them.

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -406,6 +406,9 @@ $background_image = "images/unsplash-space.jpeg";
 $custom_css = "";
 $display_footer = true;
 
+# Customized template directory: to ovverride some templates files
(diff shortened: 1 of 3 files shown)


necarnot:
Hi @coudot ,

After having upgrade from 1.7.2-1 to 1.7.3, I also integrated this PR, doing it this way :

- I'm (docker-)building it at every release, so I merged this PR
- My compose.yml contains this kind of volume binding : `- ./config/templates_custom/:/var/www/templates_custom`
- In `./config/templates_custom`, I only copied then changed the index.tpl file
- After a rebuild with the latest 1.7.3, I'm using it and everything is working fine : the customized index.tpl is used.

Thank you Clément!

PR #1052, opened 2025-03-19
Document how use env to configure LDAP TLS/SSL options

### LINKED ISSUE #1044: TLS: peer cert untrusted or revoked (0x42) - Docker v1.7.2
docker container v1.7.2 is based on debian 12 and when using it with windows AD / ldaps you'll get `TLS: peer cert untrusted or revoked (0x42)` as error message in debug docker logs, as well as `Cannot access LDAP directory ` on the GUI. 

the fix is to do a `apt update && apt install libldap-common -y` but not on the server where the docker container runs, but inside the docker container. 

to do so, open a shell inside the conatiner like `docker exec -it $CONTAINERNAME /bin/bash` and install the package as described above; then do a `docker restart $CONTAINERNAME`and it should be fine again. 

Maybe the package can be included in new docker container versions? 

Thanks.

abpai94:
Closes #1044

### DIFF
--- a/docs/config_ldap.rst
+++ b/docs/config_ldap.rst
@@ -56,6 +56,13 @@ the requirements of your server. For example:
 
    TLS_CIPHER_SUITE TLSv1+RSA
 
+Alternatively these configurations can be overriden directly in the
+``config.inc.local.php`` using the following configurations:
+
+::
+
+   putenv("LDAPTLS_REQCERT=allow");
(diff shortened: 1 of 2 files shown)


coudot:
We can't enable these new options by default:
* Setting "allow" to TLS_REQCERT will allow a bad certificate
* Setting TLS_CACERT to /etc/ssl/certs/ca-certificates.crt will work only on Debian

You can add theses options in config but they should be commented

PR #1064, opened 2025-04-11
chore: Improve login field naming

maichouni-mitek:
This is a minor visual change for the English language interface.
Some end users get confused and do not know what to write in the field.
Changing it to "Username" matches what's in the other languages.


coudot:
Thanks

PR #1067, opened 2025-04-29
Additional password checks

qrainaud:
### New password check option : check if new password contains either name or surname
- Can be enabled/disabled with `$pwd_diff_namesurname` in config
- Backend verification happens in `htdocs/change.php` after submit
- `/rest/v1/getnamesurnamefromlogin.php` api endpoint for future frontend feedback
- Frontend feedback will require ltb-project/ltb-common to be updated, as well as rest api to be enabled


qrainaud:
For frontend feedback : https://github.com/ltb-project/ltb-common/pull/71

coudot:
Hello,

why don't use `$pwd_forbidden_ldap_fields` ?
https://self-service-password.readthedocs.io/en/stable/config_ppolicy.html#forbidden-ldap-fields

qrainaud:
Hello,

I didn't notice this. The backend part of my fork is basically useless, but I believe the api as well as the frontend would still be useful as a lot of users tend to put their name or surname in their password, and the feedback message after submission isn't really clear.
And since the message is only shown after submission, they have to rewrite everything.

qrainaud:
Hello, let me know what you think.

qrainaud:
I can change it so that it checks every forbidden ldap field dynamically when the user is typing its password, and display a feedback similar to "Your password may not contain personal informations". What do you think?

coudot:
You can give a try.

I don't swear it will be included as it can be a security flaw: this would allow to discover informations from LDAP directory.

qrainaud:
For the security, I will change the api endpoint so that it validates passwords instead of returning LDAP data.
On the api side : to avoid spamming the LDAP directory, I can make a cache system so that it only fetches informations from LDAP when the login changes.

qrainaud:
Hello again, let me know what you think.

coudot:
Sorry for the lack of activity, we are not focused on SSP development for now. I keep this issue open for review

PR #1084, opened 2025-07-30
Update Japanese Translation

fu7mu4:
This service is easy to use. It seems good to us. Thanks.


fu7mu4:
I've finished this PR.

PR #1086, opened 2025-08-04
Correct time zone

mako777:
Hello,
In audit logs time is in "UTC", this PR sets time zone to the user specified "TZ" environmental variable, if there is no "TZ" set, then defaults to "UTC"
Have a nice day.


coudot:
We should maybe create a configuration parameter like `$date_timezone` which exists in Service Desk : https://service-desk.readthedocs.io/en/stable/general-parameters.html#dates

abpai94:
@mako777 thank you for your contribution, I have built upon your work to make it work for self-service-password along with adding some documentation, you can see the new PR here: #1127

PR #1090, opened 2025-08-26
Fix validation of SSH public keys

s-hamann:
I found a bug in `check_sshkey`. Key using algorithms with long names (longer than 0x1F characters) get incorrectly rejected.
One example for such an algorithm is `sk-ecdsa-sha2-nistp256@openssh.com` (0x22 characters). This results in an extra `"` in `$check`, which then obviously does not match `$algorithm`.

The first four bytes of the decoded public key specify the length of the following algorithm name.
`preg_replace` removes these four bytes *if* they are in the non-printable range. Otherwise some of the length bytes end up in `$check`.

This PR fixes this by always discarding the first four bytes and comparing the expected number of bytes.

### DIFF
--- a/lib/functions.inc.php
+++ b/lib/functions.inc.php
@@ -186,8 +186,8 @@ function check_sshkey ( $sshkey, $valid_types ) {
             return false;
         }
 
-        $ealg = base64_encode($algorithm . " AAAA");
-        $check = preg_replace('/[\x00-\x1F\x7F]/u', '', base64_decode(substr($key, 0, strlen($ealg))));
+        $raw_key = base64_decode($key);
+        $check = substr($raw_key, 4, strlen($algorithm));
         if ((string) $check !== (string) $algorithm) {
             return false;
         }


coudot:
We may find a cleaner way to check if the key is valid.

Maybe with phpseclib: https://stackoverflow.com/a/65993238

s-hamann:
Using a proper library for the purpose certainly makes sense.
In the mean time, would you consider merging this PR? It turns a flawed validity check into one that does not reject valid keys. I think that's an improvement ;)

abpai94:
Looking into this library (https://phpseclib.com/docs/publickeys) although it does provide support for the following keys:
```
ecdsa-sha2-nistp256 =>  [type] => EC
ssh-ed25519         =>  [type] => EC
ssh-rsa             =>  [type] => RSA
```

However with the following keys which require a physical security key enrollment to generate the key the library isn't valid:
```
ecdsa-sk
ed25519-sk
```

abpai94:
@s-hamann Thank you for your contribution, I have added some tests for `ecdsa-sk` and `ed25519-sk` and created a new merge request #1126 which includes your work.

PR #1095, opened 2025-10-03
Update config.inc.php smarty Version

s0ftcorn:
Im using the docker image, but even in the repo there is this line: https://github.com/ltb-project/self-service-password/blob/ae18dc85a793f927b591bc255138a775ce9e8765/conf/config.inc.php#L511

Why does it refer to smarty3? I dont know for manual installs or via apt repo, but in the docker image only /usr/share/php/smarty4/ exists. Which makes sense since in https://github.com/ltb-project/self-service-password/blob/ae18dc85a793f927b591bc255138a775ce9e8765/packaging/docker/Dockerfile#L3 smarty4 gets installed.

Changing the line to /usr/share/php/smarty4/Smarty.class.php also makes the container work instead of throwing a HTTP 500.

### DIFF
--- a/conf/config.inc.php
+++ b/conf/config.inc.php
@@ -508,7 +508,7 @@ if (file_exists (__DIR__ . '/config.inc.local.php')) {
 
 # Smarty
 if (!defined("SMARTY")) {
-    define("SMARTY", "/usr/share/php/smarty3/Smarty.class.php");
+    define("SMARTY", "/usr/share/php/smarty4/Smarty.class.php");
 }
 
 # Set preset login from HTTP header $header_name_preset_login


davidcoutadeur:
Hello,

The smarty3 path is used as a default parameter in `config.inc.php`. When installing self-service-password with rpm/deb packages, this was the better strategy, as there was no smarty4 package for a very long time.
You are encouraged to change this PATH into config.inc.**local**.php to the one of smarty4 if you have it.

In the docker container, the path in `config.inc.php` should be adapted automatically by `packaging/docker/install` script. You should give a look to why this line hasn't been correctly applied during image build.

PR #1103, opened 2025-11-20
display error logs even when $debug=false (#1102)

### LINKED ISSUE #1102: display error logs even when $debug=false
When `$debug=false`, there are no logs at all, even with critical errors that prevent the page rendering.

We should at least display error logs.

davidcoutadeur:
Fixes #1102


davidcoutadeur:
I have added the new $debug_level parameter in a new commit (see  2944c6f and https://github.com/ltb-project/white-pages/issues/178)

PR #1125, opened 2026-04-23
Show/Hide password review

### LINKED ISSUE #1099: Add an option to show/hide password in password fields
A common request we have is to add a show/hide button on password inputs like we have in a lot of products.

abpai94:
Closes #1099


coudot:
Hello @abpai94, thanks for the work, I have done some changes, mostly to avoid changing CSS file, and better use bootstrap components. Also applied to all password fields and added documentation.

PR #1133, opened 2026-05-12
Updated to latest php and alpine base images

### LINKED ISSUE #1132: Published docker image contains vulnerable apache httpd <2.4.67
With multiple critical vulnerabilities until apache httpd version 2.4.66, the docker image should be rebuilt with apache httpd 2.4.67.

abpai94:
Closes #1132


coudot:
Hello @abpai94 could explain why you need to remove opcache?

abpai94:
In php 8.5 it is preloaded by default, we have to actively choose to remove it from the build if we don't need to use it.

Source: https://www.php.net/manual/en/migration85.incompatible.php#migration85.incompatible.opcache

coudot:
Perfect, thank you

PR #1143, opened 2026-06-09
Fix call of change_password function insed multi-ldap-change script

### LINKED ISSUE #1142: multi_ldap_change.php: $directory should be $s_directory in change_password() call
## Bug description

In `scripts/multi_ldap_change.php`, the `change_password()` call on line 146 passes `$directory` which is undefined in the foreach scope, causing a fatal error.

The correct variable is `$s_directory`, initialized just above by the `switch($s_ldap_type)` block.

## Error

```php
PHP Warning:  Undefined variable $directory in /var/www/scripts/multi_ldap_change.php on line 146
PHP Fatal error:  Uncaught Error: Call to a member function computePassword() on null in /var/www/lib/functions.inc.php:94
```

## Steps to reproduce

1. Configure `$secondaries_ldap` with `ldap_type = activedirectory` in `config.inc.php`
2. Enable posthook: `$posthook = "php /var/www/scripts/multi_ldap_change.php"`
3. Trigger a password change

## Fix

```php
// line 146 - before
$result = change_password( $directory, ...

// after
$result = change_password( $s_directory, ...
```

## Version

SSP 1.8.0

mmasquelin:
$directory is undefined in the foreach scope, causing a fatal error when change_password() is called. The correct local variable is $s_directory, initialized by the switch($s_ldap_type) block above.

### DIFF
--- a/scripts/multi_ldap_change.php
+++ b/scripts/multi_ldap_change.php
@@ -143,7 +143,7 @@ foreach ($secondaries_ldap as $s_ldap) {
         # Change password
         #==============================================================================
         if ( !$result ) {
-            $result = change_password( $directory, $ldapInstance, $userdn, $newpassword, $ldap_options, $samba_mode, $samba_options, $shadow_options, $hash, $hash_options, 'manager', $oldpassword, $ldap_use_exop_passwd, $ldap_use_ppolicy_control, false, "");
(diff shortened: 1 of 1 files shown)


coudot:
Thanks!

mmasquelin:
You're welcome ! Thanks for your awesome job @coudot !

PR #1146, opened 2026-06-17
Polish the Polish language translations in lang/pl.inc.php

wsprolsztyn:
Updated Polish translations for various messages related to password management and user authentication.


coudot:
Thanks!

